PR #6, opened 2015-10-10
Cache then network recipe

TimAbraldes:


cache-then-network/index.html
```diff
@@ -0,0 +1,36 @@
...
+  </form>
+
+  <form>Network:
+    <input id='network-fail' type='checkbox'>Fail</input>
+    Delay: <input id='network-delay' type='number' value='5000'></input>ms
+  </form>
+
+  <button id='getDataButton'>Get new data</button>
+
+  <script src="./sw.js"></script>
```
digitarald: Why is the sw in the page?

TimAbraldes:
Removed WIP tag. This no longer uses a ServiceWorker at all as that isn't necessary for this demo. It shows off the Cache and Fetch APIs from the main page. We should probably create a separate demo that does live-updating using a SW that communicates with the main page via `postMessage`

cache-then-network/index.js
```diff
@@ -0,0 +1,188 @@
...
+          try {
+            handleCacheFetchCompletion(res);
+            resolve();
+          } catch (err) {
+            reject(err);
+          }
+        }, cacheDelay);
+      });
+    }).then(function() {
+      var now = window.performance.now();
```
digitarald: If you use `Date.now()` you would not need to round here. Do you see sub-millisecond results?

cache-then-network/index.html
```diff
@@ -0,0 +1,35 @@
...
+    body {
+      font-family: monospace;
+      font-size: 1rem;
+      white-space: pre-wrap;
+    }
+  </style>
+</head>
+<body>
+
+  <label>Data: </label><output id='data'></output>
```
digitarald: Consistency: " instead of ' (" use started in `head`)

cache-then-network/index.js
TimAbraldes: > Do you see sub-millisecond results? 

Not with `setTimeout` involved.

> If you use Date.now() you would not need to round here.

Changed!

cache-then-network/index.html
TimAbraldes: > Consistency: " instead of ' (" use started in head)

Changed all '"' to '''

Issue #1, opened 2015-09-22
Recipe: Test-driven ServiceWorker development


digitarald:
How to do automated tests with and without network.

lodr:
@digitarald can you explain this a little bit more, please?

darkwing:
What @delapuente said -- I don't know how this would work or what test tooling would be needed.

marco-c:
I think @digitarald meant a recipe that shows how to test an app that uses service workers (testing that it works with or without network connectivity.

vigneshshanmugam:
https://github.com/GoogleChrome/sw-unit-test-sample might help

PR #7, opened 2015-10-12
Recipe: live-flowchart

franciov:
This PR is for adding a new recipe to the cookbook: live-flowchart

This recipe provides a way to learn how to use service workers (SW) through showing [the flow diagram of SW workflow explained on the Mozilla Developer Network](https://developer.mozilla.org/en-US/docs/Web/API/Service_Worker_API/Using_Service_Workers), and logging on screen the steps taken by a real Web App running service workers.

More info about usage and compatibility in the [README file](https://github.com/franciov/serviceworker-cookbook/blob/recipe/live-flowchart/live-flowchart/README.md)

@digitarald This is the register/unregister example you mentioned. I've enhanced it a bit providing a better management of the SW workflow, a better logging with different levels (colors), adding a reload button and give the recipe a context (the flowchart). It can be used as learning tool, debug app, and I'm thinking to start from this recipe to work on #1 (test-driven service workers development) by adding the "go offline", "go online" (and "go lie-fi"?) buttons / let the SW do something useful for testing on install/activate/fetch ...


live-flowchart/app.js
```diff
@@ -0,0 +1,73 @@
+// import Logger from 'logger';
+// import SWUtil from 'sw-util';
+
+function App() {
+  var _self = this;
```
digitarald: Nit: `app` would be a better reference name.

live-flowchart/index.html
```diff
@@ -0,0 +1,34 @@
...
+		<div class="credits">
+		 	<a href="http://www.francesco.iovine.name/" target="_blank">@franciov</a>
+	 	</div>
+
+		<script src="logger.js"></script>
+		<script src="sw-util.js"></script>
+		<script src="app.js"></script>
+
+		<script>
+			new App();
```
digitarald: Move script content to JS to have it all separated for easier reading.

live-flowchart/app.js
```diff
@@ -0,0 +1,73 @@
...
+}
+
+App.prototype = {
+
+  enableCoolFeatures: function enableCoolFeatures() {
+    Logger.log('\nApp.enableCoolFeatures()');
+
+    Logger.log('Configuring service worker');
+
+    this.swUtil.registerServiceWorker(
```
digitarald: This might benefit from using promises to go hand in hand with SW's design paradigms.

digitarald:
Added some nits for the code. I like the bigger depth of this to explain the flow, but we might want to think in general how we handle the overhead of service worker code + logging without additional helpers that make understanding the recipe harder. What was your experience with that, @TimAbraldes ?

Also I'd like to get a sense for sticking to ES5 or ES6 - showing off bleeding edge APIs sounds like something that should go hand in hand with using modern language features. Thoughts? @brendandahl @marco-c

Next steps:
- Responsive to work on mobile
- :bulb: Idea: Have the flowchart build in SVG and visualize the service worker states.

live-flowchart/app.js
franciov: Good point.

live-flowchart/app.js
franciov: Actually App is a class (in ES6), shouldn't be capitalized?

franciov:
@digitarald Good points, here are some quick notes.

For helpers, they can be moved to the HTML, even if I think that providing different JS files, one for each responsibility makes the recipe more readable, especially if it's going to grow as an example app. Maybe I can add some docs and specs?

I started with ES6, worked heavenly on Firefox Nightly, but when testing Chrome Canary, old version of Firefox, and Safari (for the "SW not compatible in this browser" case) I moved back to ES5 to make sure the recipe runs everywhere. Maybe I can use a ES6 polyfill?

Love the next steps :)

marco-c:
> Also I'd like to get a sense for sticking to ES5 or ES6 - showing off bleeding edge APIs sounds like something that should go hand in hand with using modern language features. Thoughts? @brendandahl @marco-c

I do like using modern language features (and showing how to use them), but it would make recipes a little harder to follow (developers reading the recipe could get confused by language features they aren't yet comfortable with) and a little more complex and less self-contained (if we have to use a transpiler, e.g. Babel, we need to add a build step).
So, there are pros and cons, I'm unsure what's best.

TimAbraldes:
> [...] we might want to think in general how we handle the overhead of service worker code + logging without additional helpers that make understanding the recipe harder. What was your experience with that, @TimAbraldes ?

For recipes I think it's just important that someone reading the code gets a clear picture of where errors are possible, what conditions the errors represent, and what options are available for dealing with them. Whether that's a `console.log('This specific error happened')` or a `helper.displayError('This specifc error happened')`, I think it's still clear to the reader what's going on. If we start drifting a little further into abstraction and the error handler is no longer visible, then I'd start to worry.

franciov:
@digitarald I pushed a new version with:
- ES6 compatible with Firefox Nightly and Chrome Canary (I think ES6 wins?)
- new inputs to set up the SW scriptURL and scope, in order to enable more test cases
- screenshots of a few test cases I tried
- some bug fixing, mainly about checks on swRegistration.installing or swRegistration.active.
- JSDoc-style docs and more comments
- less lint errors

digitarald:
If you switch back to ES6 I need to switch the lint default ;)

franciov:
@digitarald Ups, sorry :tongue:  I thought ES6 works better if the recipe can be run on the last versions of Firefox and Chrome at least, and it does. Also, if service workers are supported, likely ES6 is supported as well. No problem to get back to ES5 in case. Thanks!

digitarald:
A subset of ES6 that has enough adoption would be OK (i.e. not `class`). `const` should be fine.

franciov:
@digitarald Yea, true, but using `class` makes the code really easy to read, and it's supported by Firefox Nightly, Chrome Canary (in strict mode), and last version of Opera (32). What it's not yet ready to be used, that I noticed, is `let`, `import`, and `includes` called on a `string`.

digitarald:
The issue is that `class` might be in Nightly but not ride the same train as service workers.

Should we run code through babel with classes and other close-but-not-yet es6 features as the only transforms?

franciov:
I've just pushed a version without `class`: it seems that arrow functions are not allowed as well? Looking at the compatibility table on MDN, arrow functions are quite supported on desktop, but not widely supported on mobile browsers: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions#Browser_compatibility

digitarald:
Arrow function are only allowed in strict mode, might that be the issue? That Chrome/Android doesn't support them is of course a surprise. So no fat arrows for the cookbook :(

I guess we really need babel at one point to show es5 and es6 at the same time.

franciov:
Success \o/ @digitarald Basically now it's ES5 powered with ES6 Promises. Also here for quick testing: https://franciov.github.io/serviceworker-cookbook/live-flowchart/

live-flowchart/sw.js
```diff
@@ -0,0 +1,44 @@
+/*
+ * This is a service worker
+ */
+
+console.debug('\nsw.js: this is a service worker');
```
digitarald: Why the extra \n in front of all messages?

live-flowchart/sw-util.js
```diff
@@ -0,0 +1,178 @@
...
+  );
+};
+
+/*
+ * @method unregisterServiceWorker
+ * Unregister the active service worker
+ * @returns {Promise}
+ */
+SWUtil.prototype.unregisterServiceWorker = function unregisterServiceWorker() {
+  Logger.log('\nSWUtil.unregisterServiceWorker()');
```
digitarald: Is the line break separating groups of logs or something similar?

live-flowchart/logger.js
```diff
@@ -0,0 +1,116 @@
...
+    var newLine;
+
+    logMessage = document.createElement('div');
+
+    if (message) {
+      // add message to the HTML node
+      logMessage.innerHTML = message;
+
+      // new like check
+      if (Object.prototype.toString.call(message) === '[object String]'
```
digitarald: Why not `typeof message === "string"``?

live-flowchart/sw.js
```diff
@@ -0,0 +1,44 @@
+/*
+ * This is a service worker
+ */
```
digitarald: More comments will help to make the code the documentation for the project. See https://serviceworker-cookbook.herokuapp.com/docs/auto-claim/service-worker.html for documentation output.

live-flowchart/sw.js
franciov: `\n` is to write a new line. In sw.js to separate each state of the SW execution (`init`, `oninstall`, `onactivate`, `onfetch`), to make it easier to read. I can do the same with a `console.debug('\n')` or a `console.debug('---')` before instead.

live-flowchart/sw-util.js
franciov: Yep, both in the browser log and in the HTML log. I can do the same with a `Logger.log('\n')` or a `Logger.log('---')` before instead, or alternatively remove the new lines. Does it break the doc generation?

live-flowchart/logger.js
franciov: Yea, `typeof message === "string"` is way more readable, and do the same job in this context. Thanks.

franciov:
@digitarald 
- added a new more clean `Logger.newSection()` method, instead of using `\n`
- added more comments on `sw.js` for generating documentation with docco

It seems that docco process only inline comments, is it decision / by design or it can be customised? It would be cool if docco could generate documentation based on comments like this:
```
/*
 * @method registerServiceWorker
 * Register a service worker
 * @param {String} path The script URL of the service worker to be registered
 * @param {String} scope (optional) The scope of the service worker to be registered
 * @returns {Promise}
 */
```
It's ok to use only inline comments in case.

Mind also that I had work around a problem with docco (https://github.com/mozilla/serviceworker-cookbook/issues/14) after syncing the forked repo.

digitarald:
> It would be cool if docco could generate documentation based on comments like this:

We currently don't process jsdocs and I'd like to hear ideas on why it would make sense. My point would be that the recipe should not focus on creating a library with a reusable interface (what jsdoc documents) but to demo the actual Web APIs as much as possible. When we get too many mini-libraries in the recipes developers will copy and re-use them in their own projects even though they weren't intended for that purpose. If we have libraries that people should use, we could add specific recipes for them.

live-flowchart/sw.js
```diff
@@ -0,0 +1,47 @@
+console.debug('sw.js: this is a service worker');
+
+// This is the basic architecture of a service worker, as described here:
+// https://developer.mozilla.org/en-US/docs/Web/API/Service_Worker_API/Using_Service_Workers#Basic_architecture
+
+// With service workers, the following steps are generally observed for basic set up:
+
+// (1) The service worker URL is fetched and registered via serviceWorker.register().
```
digitarald: You can use markdown in those comments to get even prettier docco output.

franciov:
@digitarald I made the comments Docco-compliant and added some markup. I tried to make it look nice, linked MDN docs. 

For mini-libraries, I think that using simple js helpers (`SWUtil`, `Logger`) makes the code more readable, extendable, and limit the copy-and-paste of lines of code: the developer is somehow pushed to understand what it's going on inside methods such `SWUtil.registerServiceWorker`. In the documentation I focused on describing the service workers basic architecture and workflow, rather than documenting the helpers. And used the same approach with `Logger`, avoiding printing HTML/browser logs about the helpers. The flip-side, as you mentioned, is that in some cases developers could re-use and extend the helper, treating it as a best-practice module that abstracts the use of service workers, and this might not be the case. To avoid this I can transform all object methods in global functions.

digitarald:
Thank you @franciov, we really got a showcase recipe with this one !

PR #8, opened 2015-10-12
eslint rules

digitarald:
Removing strict (assumes babel) and id-length (for loops are not the same without i)

More to add/remove?

r? @TimAbraldes @franciov

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -2,6 +2,8 @@
   "extends": "airbnb/base",
   "rules": {
     "no-console": 0,
-    "no-use-before-define": 0
+    "no-use-before-define": 0,
+    "strict": 0,
+    "id-length": 0
   }
 }
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,9 +1,6 @@
 {
-  "extends": "airbnb/base",
+  "extends": "eslint-config-airbnb-es5",
   "rules": {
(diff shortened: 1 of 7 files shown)


franciov:
@digitarald If we go for ES5 (in lint), I would remove `no-var` as well: `let` and `const` might not be supported.

Also, "do not use `new` for side effects" sometimes is not an error, maybe a warning? Like when calling a `new App()` or `new Game()`, and you don't want to define a global `var`/`let`/`const`: https://jslinterrors.com/do-not-use-new-for-side-effects

The lint errors on the `live-flowchart` recipe:
```
/Users/franciov/Projects/serviceworker-cookbook/live-flowchart/app.js
   4:3  error  Unexpected var, use let or const instead  no-var
  75:1  error  Do not use 'new' for side effects         no-new

/Users/franciov/Projects/serviceworker-cookbook/live-flowchart/logger.js
  2:1  error  Unexpected var, use let or const instead  no-var
  7:5  error  Unexpected var, use let or const instead  no-var
  8:5  error  Unexpected var, use let or const instead  no-var
```

digitarald:
Lets switch to a lint that supports both for now.

I am leaning towards using es6 exclusively as most sw examples out there use es6 syntax (arrow functions).

Issue #2, opened 2015-09-22
Recipe: Load list of cachable files from JSON file


digitarald:
*aka "Poor Man's AppCache"*

https://jakearchibald.com/2014/offline-cookbook/#comment-2189777576

https://github.com/mozilla/oghliner/issues/10

PR #28, opened 2015-10-26
Issue 2 - Add basic demo for loading cachable files from JSON

darkwing:
Very basic which seems to fit issue #2.  Let me know how I can improve it, please!


darkwing:
Need to update lint, it seems.  On it.

darkwing:
I believe this is ready for a look.  Any feedback is very welcome!

digitarald:
@delapuente can you also have a look?

- The demo doesn't invalidate the old cache. When the JSON updates it would not remove the outdated files. The JSON file could either have a version number for the cache or hashes for the files to only evict and download changed files. *Maybe that could be a second example (`json-cache-versioned`) though, to not complicate this one*
- Could we have more random pictures ;)

darkwing:
@digitarald : Thank you for looking at this so quickly.

1.  Where can I find a good example / writeup on the cache-busting?

2.  I'll find some better imagery, it was the least of my concerns :)

darkwing:
I actually agree with the idea that we make a second recipe with cache-busting;  it would keep this one dead simple and that receipt could be focused *primarily* on a good cache-busting technique...something I'd love to see right now :+1: 

Looking forward to @delapuente 's feedback.

json-cache/index.html
```diff
@@ -0,0 +1,34 @@
...
+
+    img {
+      max-width: 200px;
+      display: inline-block;
+    }
+  </style>
+</head>
+<body>
+
+  <p>This demo uses a service worker which loads a JSON file representing files to be cached by the Service Worker.  Once the JSON file is loaded and parsed, files are placed into the cache vvia the Service Worker.</p>
```
delapuente: s/vvia/via/

json-cache/service-worker.js
```diff
@@ -0,0 +1,43 @@
+// [Working example](/serviceworker-cookbook/json-cache/).
+
+var CACHE_NAME = 'dependencies-cache';
+
+self.addEventListener('install', function(event) {
```
delapuente: Although is crystal clear for me, I think the code is undercommented for a recipe. Moreover if we want to add docco generated documentation. I think it deserves some explanations about the rationale behind each decission.

json-cache/service-worker.js
```diff
@@ -0,0 +1,43 @@
...
+  event.respondWith(
+    caches.match(event.request)
+      .then(function(response) {
+        // Cache hit - return response
+        if (response) {
+          console.log('[fetch] Returning from cache: ', event.request.url);
+          return response;
+        }
+
+        return fetch(event.request);
```
delapuente: I think it could be simpler to do
```js
return response || fetch(event.request);
```
And it's emphasyzes the fact that `fetch()` is like a fallback.

json-cache/service-worker.js
```diff
@@ -0,0 +1,43 @@
...
+      }
+    )
+  );
+});
+
+self.addEventListener('activate', function() {
+  console.log('[activate] Activating service worker!');
```
delapuente: I'm not sure about simply logging activation, maybe it's noisy but we can clarify in a comment: `// Simply loggin activation`.

json-cache/service-worker.js
```diff
@@ -0,0 +1,43 @@
...
+  event.waitUntil(
+    caches.open(CACHE_NAME)
+      .then(function(cache) {
+        console.log('[install] Caches opened');
+
+        return fetch('files-to-cache.json').then(function(response) {
+          return response.json();
+        }).then(function(files) {
+          console.log('[install] Adding files from JSON file: ', files);
+          return cache.addAll(files);
```
delapuente: It would be awesome if we could recognise the ServiceWorker API in the generated docco to provide the links to MDN when clicking on them, enhancement in issue #33. I will figure out.

json-cache/service-worker.js
darkwing: I've added more explanation as requested.

json-cache/service-worker.js
```diff
@@ -0,0 +1,53 @@
+// [Working example](/serviceworker-cookbook/json-cache/).
+
+var CACHE_NAME = 'dependencies-cache';
+
+self.addEventListener('install', function(event) {
+  // Perform the install step:
```
delapuente: Remember you can use markdown inside docco so you could replace `-` with `*`. ;)

darkwing: I didn't know...I shall update :+1:

json-cache/service-worker.js
```diff
@@ -0,0 +1,53 @@
...
+  // - Load a JSON file from server
+  // - Parse as JSON
+  // - Add files to the cache list
+
+  event.waitUntil(
+    caches.open(CACHE_NAME)
+      .then(function(cache) {
+        // With the cache opened, load a JSON file containing an array of files to be cached
+        return fetch('files-to-cache.json').then(function(response) {
+          // Once the contents are loaded, convert the raw text to JSON
```
delapuente: To be precise, I would reword this as `convert the raw text to a JavaScript object` as JSON **is** actually raw text.

darkwing:
This should be ready for review again

json-cache/service-worker.js
```diff
@@ -0,0 +1,53 @@
...
+        console.log('[fetch] Returning from server: ', event.request.url);
+        return fetch(event.request);
+      }
+    )
+  );
+});
+
+self.addEventListener('activate', function() {
+  // Message to simply show the lifecycle flow
+  console.log('[activate] Activating service worker!');
```
delapuente: Forgot to mention, but you should to call [calim()](https://developer.mozilla.org/en-US/docs/Web/API/Clients/claim) here.

darkwing:
I've added a `claim()` call to this recipe; Firefox shows the following flow upon initial page hit: 

recipeflow

Then upon request to the page while offline:

recipeflowoffline

Please confirm that this is correct.  I still don't completely understand why `claim` is needed or what exactly it does.  It's also unclear to me if `addAll` makes the fetches to the files or if a refresh is needed or how that sequence of events goes.

Service Workers are difficult to completely understand :/

digitarald:
This should help, if its not clear, I fix it: https://serviceworker-cookbook.herokuapp.com/docs/auto-claim/service-worker.html

Normally when a worker is registered it will trigger `oninstall`.  It will only `activate` though when the page is refreshed, at which point it calls `onactivate` and takes over control the page (aka claim and firing `oncontrollerchange` on the client). This way the page still runs on the previous service worker version when a new version is found until the page is refreshed.

With manual claim, you a) skip directly from install to activate and b) claim the page directly in activate.

json-cache/service-worker.js
```diff
@@ -0,0 +1,60 @@
...
+          return response.json();
+        }).then(function(files) {
+          // Use cache.addAll just as you would a hardcoded array of items
+          console.log('[install] Adding files from JSON file: ', files);
+          return cache.addAll(files);
+        });
+      })
+      .then(function() {
+        // Message to simply show the lifecycle flow
+        console.log('[install] All required resources have been cached; the Service Worker was successfully installed!');
```
digitarald: You are missing a `return self.skipWaiting();` here to force activation.

darkwing: Added!

Issue #5, opened 2015-10-09
Recipe: Offline status in frontend


digitarald:
Per @wfwalker and his project https://github.com/wfwalker/mc-choir/ , it would be great to have an indication when the project is cached/*offlined* (`Dear user, you can now safely cut the cable!`). This is probably a mix of `claim` and listening to `controllerchange` in the frontend.

darkwing:
I'd like to have a go at this, if that's cool.

darkwing:
A few questions I've encountered as I've started this:

[ ] - Do you envision this using two service workers -- one for static assets for layout and then another for other media?

[ ] - Can we detect assets included within the HTML document itself, including `<img>`, `<link>`, via service worker?

[ ] - How does `controllerchange` factor into this?

digitarald:
1. You can only have one active service worker
2. The assets are defined [within the service worker](https://github.com/wfwalker/mc-choir/blob/gh-pages/offline-worker.js#L30). The sw will only `activate` when all assets are cached
3. To get `controllerchange` the sw needs to claim the page, which by default only happens on refresh but which can be forced by calling `.claim()` (see [auto-claim](https://github.com/mozilla/serviceworker-cookbook/tree/master/auto-claim)).

darkwing:
1.  Good to know
2.  Right, though they can be added as we go along with:  

```
self.addEventListener('fetch', function(event) {

  console.log('fetch!  ', event.request.url);

  event.respondWith(
    caches.match(event.request)
      .then(function(response) {
        // Cache hit - return response
        if (response) {
          console.log('Returning from ServiceWorker cache:  ', event.request.url);
          return response;
        }

        var fetchClone = event.request.clone();

        return fetch(event.request).then(function(response) {
          // Check if we received a valid response
          if(!response || response.status !== 200 || response.type !== 'basic') {
            return response;
          }

          var responseToCache = response.clone();

          caches.open(CACHE_NAME).then(function(cache) {
            console.log('Putting into SW cache: ', event.request.url);
            cache.put(event.request, responseToCache);
          });

          return response;
        });
      }
    )
  );
});
```

Right?

3.  So I'm confused as to the scope of this recipe.  Are we enhancing Bill's app, moving it over or...?  At face value this recipe seems incredibly simple but you've mentioned `controllerchange` and `claim()` so now I think I need more insight as to what's desired.

darkwing:
This is set for PR review

PR #34, opened 2015-10-28
Issue 5 - "Offline Status" recipe

darkwing:
This recipe covers issue #5.  To Do:

- [x] Cleanup event listeners once we notify the user the app is ready for offline
- [x] Write intro text on `index.html` explaining what this recipe does
- [x] Get rid of gross styling
- [x] Make the example actually do something ?  :D
- [x] Lint


offline-status/index.js
```diff
@@ -0,0 +1,32 @@
...
+navigator.serviceWorker.register('service-worker.js', {
+  scope: '.'
+}).then(function(registration) {
+  console.log('The service worker has been registered ', registration);
+});
+
+// Listen for claiming of our ServiceWorker
+navigator.serviceWorker.addEventListener('controllerchange', function(event) {
+  console.log('[controllerchange] A "controllerchange" event has happened within navigator.serviceWorker: ', event);
+
```
darkwing: Should we do some sort of `scriptURL` validation here to ensure it's the correct service worker?

digitarald: We would assume most people have one sw per scope right now. There shouldn't be conflicts.

offline-status/service-worker.js
```diff
@@ -0,0 +1,56 @@
...
+
+// Files required to make this app work offline
+var REQUIRED_FILES = [
+  'assets/mdn-logo.png',
+  'assets/player.js',
+  'assets/style.css',
+  'assets/users.json',
+  'index.html',
+  'index.js',
+  'service-worker.js'
```
digitarald: Not needed.

offline-status/index.js
darkwing: :+1:

darkwing:
OK, this should be ready for review!

darkwing:
Per @digitarald 's request, I've removed use of the Notification API and instead show a "toaster" notification.

digitarald:
What about showing the notification when the permission was already given, this would be an interesting workaround around randomly showing permission popups without context. The domain could have asked before to show notifications.

Issue #25, opened 2015-10-20
Document technical specs of running app


darkwing:
I installed a few different versions of Node.js before finding a version that supported `String.prototype.startsWith`.  Having a tech spec doc would be helpful for working within this repo.

digitarald:
I am just spoiled by babel adding polyfills ;) . We should remove node 4 dependencies to make it more accessible.

darkwing:
Going to close this as we've not seen other problems here.

Issue #26, opened 2015-10-20
Make build process work with older node versions.


digitarald:
@darkwing stumbled over `String::startsWith` in #25 and there might be other references.

darkwing:
Please assign to me as I have a PR in

PR #36, opened 2015-10-29
Issue 25 - Allow running server on 'older' node

darkwing:
Fixes issue #25 ; server will now start on 0.12.7, for example.


marco-c:
This actually fixed #26.

PR #37, opened 2015-10-29
Add recipe for sending message to other clients.

brendandahl:


brendandahl:
source.id is currently broken in chrome, tracking in https://code.google.com/p/chromium/issues/detail?id=549346

message-relay/service-worker.js
```diff
@@ -0,0 +1,23 @@
...
+      });
+    });
+  });
+});
+
+// Immediately claim any new clients. This is not needed to send messages, but
+// makes for a better demo experience since the user does not need to refresh.
+// A more complete example of this given in the immediate-claim recipe.
+self.addEventListener('activate', function() {
+  self.clients.claim();
```
digitarald: Does this work as expected without the `skipWaiting` in `install`? https://github.com/mozilla/serviceworker-cookbook/blob/master/immediate-claim/service-worker.js#L19

brendandahl: On first install there are no waiting service workers, so it won't be put in the waiting state.

PR #38, opened 2015-11-01
Adding cache from zip recipe

### LINKED ISSUE #23: Recipe: download a zip and cache its contents
With the use of 3rd party libraries such as [zip.js](https://gildas-lormeau.github.io/zip.js/), the Service Worker can use other workers to decompress a zip in the client and use it as source for caching off-line content. For instance, [downloading GitHub provided zip for a branch](http://stackoverflow.com/questions/13636559/how-to-download-zip-from-github-for-a-particular-commit-sha#answer-13636954).

delapuente:
This fixes #23 
@digitarald do you mind to take a look?


delapuente:
Aaagh, I missed to check if it works on Nightly. It is not working because [FileReader is not exposed in ServiceWorkers](https://bugzilla.mozilla.org/show_bug.cgi?id=901097) so the ServiceWorker can not read a blob. I blocked the [shipping](https://bugzilla.mozilla.org/show_bug.cgi?id=1059784) with that bug as well.

digitarald:
@brendandahl would you know a workaround?

delapuente:
As discussed yesterday in the daily, we could perform the unzip in _content_, then passing the results back to the service worker. We could implement a polyfill to cover this functionality but not without introducing noise into the recipe. I'm going to rebase the changes.

Another possibility is to perform all the installation process in _content_ as fetch and offline cache are exposed to it and then use service workers to simply serve what is in the cache. I did not use this approach because fetching and caching are use cases closely tied to the `install` event for a service worker, cache from zip is not very different to the `addAll()` method from offline caches but it reduces the overhead introduced by making each request.

@digitarald @brendandahl  What do you think? Do you prefer I move the caching process to _content_?

delapuente:
@darkwing do you want to take a look at this other recipe?
@digitarald do we merge this recipe even if it's not compatible yet with Nightly?

cache-from-zip/worker.js
```diff
@@ -0,0 +1,94 @@
...
+
+        return cache.put(location, response);
+      }).then(fulfill, reject);
+    });
+  });
+}
+
+// Return the location for each entry.
+function getLocation(filename) {
+  var name = filename || '';
```
darkwing: We probably don't need this var, can probably just do `filename || ''` on the next line

cache-from-zip/index.html
```diff
@@ -0,0 +1,43 @@
...
+    We are going to simulate installing and uninstalling a web application
+    in the browser. To do that, enable logs from service workers to be shown
+    and click on install button.
+    <button id="install">Install</button>
+  </section>
+
+  <section id="images" hidden>
+    Choose from one the images to be loaded: <select>
+      <option value="imgs/a.jpeg">Collaboration</option>
+      <option value="imgs/b.jpeg">Lonelines</option>
```
darkwing: "Loneliness"

cache-from-zip/index.html
```diff
@@ -0,0 +1,43 @@
...
+      <option value="imgs/c.jpeg">Breads</option>
```
darkwing: "Bread" (it's plural in English)  :)

darkwing:
I think we could add a message about Nightly not working and point to the bug number.

Also, can you provide a super brief explanation of the ZIP coding?  Why are we commuting the images separately?  Does unzip not really work?

cache-from-zip/worker.js
delapuente: Nice catch!

cache-from-zip/index.html
delapuente: Good to know.

delapuente:
> I think we could add a message about Nightly not working and point to the bug number.

Fair enough.

> Also, can you provide a super brief explanation of the ZIP coding?

What do you refer with the "ZIP coding"?

digitarald:
This lib seems to allow reading from text: https://gildas-lormeau.github.io/zip.js/core-api.html#zip-reading

delapuente:
@digitarald now is working fine on nightly? do you think we can merge?

PR #39, opened 2015-11-01
Allowing copy folders inside recipes but restricting docco documentat…

### LINKED ISSUE #14: Bug: Docco does not generate documentation if a recipe contains a folder
It seems that docco, the tool for generating documentation when running `gulp build-docs`, does not work properly if a recipe contains a folder: the js files after the folder, in a lexical order, are not processed.

I had to remove the folder in my recipe to make it work: https://github.com/mozilla/serviceworker-cookbook/commit/db6567d83ea9d78fa9e7abfd76880cb4db301e78

```
$ npm -v
2.14.7
$ node -v
v4.2.1
```

delapuente:
…ion to JS for the first level
Fixes #14 
@digitarald do you mind to take a look?

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -9,7 +9,10 @@ var recipes = glob.sync('./!(dist|node_modules)/').map(function toBase(dir) {
   return path.basename(dir);
 });
 var srcRecipes = recipes.map(function makePath(name) {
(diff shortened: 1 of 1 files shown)


digitarald:
@darkwing do we still need this with #40 being fixed?

PR #40, opened 2015-11-02
Move assets to main level of examples

darkwing:
Having an `assets` folder is somehow breaking `gulp start-server` so I've removed the assets folder and put all required items at the main line


digitarald:
Agreed, better to have a flat structure for each recipe as a constrain for simplicity

PR #48, opened 2015-11-04
Simple push notification recipe

marco-c:


marco-c:
To support Chrome we need an API key. I have one that we could use, @digitarald could you add it to heroku?

marco-c:
> I think the start_url is relative to the manifest.json path, this should this be ./index.html (not 100% sure).

Yes, I've read the specs and this seems the case (https://w3c.github.io/manifest/#start_url-member).

PR #49, opened 2015-11-05
Push notification with subscribe and unsubscribe

zalun:


zalun:
There is still a need to react to unsubscribe from outside of the page

marco-c:
I filed https://bugzilla.mozilla.org/show_bug.cgi?id=1221965, which should make it easier to test push subsrciption changes.

zalun:
Recipe should react to cancelled (via notification UI) and expired push subscriptions
1. server should remove subscription if endpoint doesn't exist (server responds with 4xx error)
2. front-end should change from subscribed to unsubscribed (if cancelled) or reissue subscription (if expired)

zalun:
Because of a weak storage this recipe fails when server is restarted and the subscription remains in push service. It shows that user is subscribed to notifications, but notifications do not show. Consider if handling this use case should be a part of the recipe or simply add this info to readme.

zalun:
pushsubscriptionchanged is an event fired only if subscription is expired. This recipe should act if this happens

push-subscription-management/Readme.md
```diff
@@ -0,0 +1,29 @@
...
+
+Unsubscribe
+-----------
+After successful unsubscription (index.js::pushSubscription.unsubscribe) client sends a post request to application server to unregister the subscription. Server is no longer sending notification.
+
+Subscription expired
+--------------------
+Service worker is watching for the *pushsubscriptionchange* event and resubscribes to the push service.
+
+Not in recipe
```
marco-c: This is now implemented, with the pushsubscriptionchange event handler.

push-subscription-management/server.js
```diff
@@ -0,0 +1,61 @@
+var webPush = require('web-push');
+// collection of endpoints stored in memory (in production it should be
+// a database)
+var subscriptions = [];
```
marco-c: Could we link the subscription to a particular user? Instead of storing only the endpoints, you could store a user ID and its endpoint. This way when the subscription changes and the client notifies the server, the server can overwrite the user's endpoint.

zalun: That would add some complication, but it would be a workaround to https://bugzilla.mozilla.org/show_bug.cgi?id=1222428

marco-c: True, it would be a little more complicated. Let's see what @digitarald thinks.

digitarald: We could let the user enter a username as a "login"? But maybe that is a different recipe, where the push subscription is bound to the login status and has to be unsubscribed when the user logs out.

zalun: this is more a hashtag than a user - just so a client is identified in the server

PR #52, opened 2015-11-05
Push notification with clients management recipe

marco-c:
This currently doesn't completely work in Chrome.
Indeed, `clients.openWindow('./push-clients/index.html')` opens `http://localhost:3003/push-clients/index.html` in Firefox, but `http://localhost:3003/push-clients/push-clients/index.html` in Chrome.

https://bugzilla.mozilla.org/show_bug.cgi?id=1222097

### DIFF
--- a/push-clients/index.html
+++ b/push-clients/index.html
@@ -1,0 +1,30 @@
+<!doctype html>
+<html>
+<head>
+  <meta charset="utf-8">
+  <title>Push Notification with clients management - ServiceWorker Cookbook</title>
+  <meta name="viewport" content="width=device-width, initial-scale=1">
+  <style>
+    body {
+      font-family: monospace;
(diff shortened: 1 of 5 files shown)


digitarald:
Sounds like Chrome works as expected as the url should relative to the service worker scope? Did you file a bug, @marco-c ?

marco-c:
> Sounds like Chrome works as expected as the url should relative to the service worker scope? Did you file a bug, @marco-c ?

Yes, I mentioned it in the first comment ;)

marco-c:
I've just CCed you on the bug.

marco-c:
The bug should be fixed soon. I've updated the recipe to follow the specs.

darkwing:
I can confirm it's acting as @marco-c has found.

PR #55, opened 2015-11-06
Rich push notification recipe

marco-c:
This works in Chrome, doesn't work in Firefox (https://bugzilla.mozilla.org/show_bug.cgi?id=1222476).


darkwing:
For some reason I'm getting a 404 when the register happens, but it may just be my setup.

PR #56, opened 2015-11-09
Add README.md files for my demos

darkwing:
Important I get these in while still fresh in my mind.


digitarald:
Nicely done, good best practices. I wonder how we could add more technical depth.

PR #58, opened 2015-11-11
fetching images all modes and sources

zalun:


darkwing:
Seeing `event.request.url.contains is not a function` a few times before and after install -- is that desired?

zalun:
I don't see it - Nightly?

PR #63, opened 2015-11-14
WIP: Cache CSS

### LINKED ISSUE #9: Recipe: Cache CSS and fonts
1. Serve stale files for CSS and fonts on load
2. Fetch new version into cache (on install or on fetch?)
  - Fonts probably didn't update but just CSS, should the service worker only invalidate changed files?
3. Refresh CSS in client when updates are available

Similar to https://github.com/phamann/embrace-the-network/tree/master/src/stale-while-revalidate

TimAbraldes:
Will eventually close #9

Hopefully the text in the page and in the comments is descriptive enough to explain what's going on here. The one interesting bit that may not have been explained there is that the SW pretends to do a network fetch for CSS resources rather than actually performing the fetch. That way we don't need any server-side logic; the user can control whether an "update" has been pushed to the "server" with the push of a button on the main page.

I decided not to automatically refresh the page because it would be harder to explain what the user is seeing in that case. The goal of this recipe, after all, is to be a learning resource.

WIP so far because this completely fails in Firefox Dev Edition. Works perfectly in Chrome.


update-css/README.md
```diff
@@ -0,0 +1 @@
+# Recipe: Cache and update CSS
```
delapuente: Believe it or not, most of the problem is here. The recipe is illustrating at least, two techniques (messaging and using cache as database) non related with #9 . Defining the use case and the solution overview help you to keep the recipe focused in what you want to solve.

In #9 we are talking about the problem of caching **automatically** large files (fonts) and almost long term immutable (images and CSS) assets. This should lead to better loading times, for instance (and we could demonstrate comparing first load with the subsequent ones) and a way to manually wipe out the cache to repeat the experiment.

The solution would be to perform a generic fetch strategy that detects CSS, image and font fetching, and try to match the cache, then fetch the network to obtain a fresh response for the next serving.

Although _recipe_ could point to a mixture of techniques to solve a problem, I think recipes should not show more than one set of features (CSS case is to inspect request properties to take decisions for fetching or not) at a time.

Messaging and the cache-as-db trick **are great** (I use the latter in offliner) but are not mainly related with caching large / semi static files. More even when, for messaging, the SW only acts like a pass through channel for messages and as a message handler; furthermore, you will realize that it is not as simple to get a working solution for Nightly and Chrome.

So, summarizing, my humble opinion is: write the use case, write the overview for the solution and design an experiment to make the user realize about the value of your proposal.

Then try to stick with it and introduce as fewer noise as possible.

Let me know what you think.

delapuente: Of course you can split this recipe and illustrate messaging and cache-as-db in other two different recipes :)

update-css/sw.js
```diff
@@ -0,0 +1,108 @@
+var cacheName = 'update-css-2015.1015.1202';
+
+self.addEventListener('install', function(ev) {
```
darkwing: We've been using `event` instead of `ev` everywhere else, may be best to be consistent.

update-css/main.js
```diff
@@ -0,0 +1,50 @@
+var cacheName = 'update-css-2015.1015.1202';
+
+window.navigator.serviceWorker.register('sw.js');
```
darkwing: nit: We've just been using `navigator.serviceWorker.register('sw.js')`

TimAbraldes:
Hopefully I'll respond to each point, but not in order.

> In #9 we are talking about the problem of caching **automatically** large files (fonts) and almost long term immutable (images and CSS) assets. This should lead to better loading times, for instance (and we could demonstrate comparing first load with the subsequent ones) and a way to manually wipe out the cache to repeat the experiment.

It sounds like you're envisioning a page that loads a bunch of big resources so that the page takes a long time to load, then on subsequent loads it uses cached versions of the resources so that the load time is much shorter. I could certainly add this to the recipe, and it sounds like a cool demo, but it's not part of the task I set out to accomplish.

Quoting the original comment that was written when opening #9:

> 1. Serve stale files for CSS and fonts on load
> 2. Fetch new version into cache (on install or on fetch?)
>  - Fonts probably didn't update but just CSS, should the service worker only invalidate changed files?
> 3. Refresh CSS in client when updates are available

I set out to show one way _how_ to do this, not _why_ to do this.

> The solution would be to perform a generic fetch strategy that detects CSS, image and font fetching, and try to match the cache, then fetch the network to obtain a fresh response for the next serving.

This is exactly what the recipe attempts to demonstrate, though I left out the "detection" part and simply fetched/cached a single, known CSS file. If this is something we want, I can update the `FetchEvent` handler in the Service Worker to cache all CSS, image, and font fetches, and add some CSS, images, and fonts to the page to be fetched.

> Although _recipe_ could point to a mixture of techniques to solve a problem, I think recipes should not show more than one set of features (CSS case is to inspect request properties to take decisions for fetching or not) at a time.
> The recipe is illustrating at least, two techniques (messaging and using cache as database) non related with #9 
> Messaging and the cache-as-db trick **are great** (I use the latter in offliner) but are not mainly related with caching large / semi static files. More even when, for messaging, the SW only acts like a pass through channel for messages and as a message handler; furthermore, you will realize that it is not as simple to get a working solution for Nightly and Chrome.

Item number 3 in the post I quoted above is "Refresh CSS in client when updates are available" and the later discussion morphed that a little bit to something like "at least notify the page when an update is available." To accomplish this, the Service Worker _must_ communicate with the page and `postMessage` is the most appropriate way to accomplish that.

The cache-as-db trick is an implementation detail: The Service Worker in a real situation would always be requesting 'style.css.' I needed to simulate a request that gets a new response without actually changing anything on a server, and I wanted the page to be able to control when an "updated" version is available. I created two other CSS files, style-1.css and style-2.css, and I cache the name of whichever one is "current." The page can change which one is considered "current," and the Service Worker can query that information from the cache. So this example _uses_ cache-as-db. It's not intended to be exemplifying that technique, so maybe I can write more comments and rearrange the code to make this clearer.

update-css/sw.js
TimAbraldes: > We've been using event instead of ev everywhere else, may be best to be consistent.

Renamed in 01e0fd0

TimAbraldes:
> [...] you will realize that it is not as simple to get a working solution for Nightly and Chrome

Could you elaborate on this?

update-css/main.js
TimAbraldes: > nit: We've just been using `navigator.serviceWorker.register('sw.js')`

Changed in 7e55a4d

TimAbraldes:
I've been expanding the example here to include more CSS files and fonts, and I've been working on a solution that avoids the scenario that a client sees a partial update (i.e. some older CSS files mixed with newer ones).

So far I'm working on these changes to the PR:

- Incorporate a 'current.json' file that lists each resource we want to cache along with something (a hash or a last-updated time perhaps) to compare for updates (as in @digitarald's comment in #9). There will be a mechanism on the page for switching which 'current.json' file is fetched by the Service Worker.
- In the Service Worker `FetchEvent` handler, we will simply return cached data. We will also kick off the validate-and-maybe-update logic.
- Validate-and-maybe-update:
  - Get current.json
  - Compare the entries in current.json with the files in the cache. If any have been updated (decided by last update time or hash or some other metric), start downloads for each file that needs updating. For each file that completes downloading, cache it in a separate, temporary cache. Once all downloads are complete, unregister the current Service Worker and `postMessage` to all clients notifying them that an update is available.
- In the Service Worker `ActivateEvent` handler, we will clobber the cache with the entries in the temporary cache, then delete the temporary cache

update-css/sw.js
```diff
@@ -0,0 +1,108 @@
...
+          cache.put('style.css', response);
+        }
+      });
+    });
+  }));
+});
+
+self.addEventListener('activate', function() {
+  if (self.clients && self.clients.claim) {
+    self.clients.claim();
```
delapuente: It could be possible (but very rare) for the UA to kill the service worker before finishing claiming. To ensure not to do this:
```js
event.waitUntil(self.clients.claim());
```
In the other hand, both browsers shipping sw already implements `self.clients` and `self.clients.claim` so you can remove the `if (...)` check.

update-css/sw.js
```diff
@@ -0,0 +1,108 @@
...
+
+  // Respond with the cached version of 'style.css' after initiating
+  // a network fetch to update our cache if necessary.
+  event.respondWith(caches.open(cacheName).then(function(cache) {
+    // Perform a network fetch of 'style.css' to see if there's an updated
+    // version available. If there is, replace our cached version with the
+    // updated one and use `postMessage` to notify the page that there
+    // is updated CSS available.
+    simulateStyleCSSNetworkFetch().then(function(response) {
+      // In real production code, we would use smarter heuristics to
```
delapuente: AFAIK, the express server includes an [Etag header](https://en.wikipedia.org/wiki/HTTP_ETag), you can use this if you want.

update-css/sw.js
```diff
@@ -0,0 +1,108 @@
...
+      // determine if the CSS file that we requested has actually been
+      // updated.
+      if (response && response.text) {
+        var clone = response.clone();
+        clone.text().then(function(networkText) {
+          cache.match('style.css').then(function(cachedResponse) {
+            if (cachedResponse && cachedResponse.text) {
```
delapuente: Again, I think all the UA implements `.text()` we should avoid these checks in the recipes to keep code simple. Wdyt?

delapuente: I'm referring to `cacheResponse.text`, the first check is needed to distinguish from `undefined`.

update-css/sw.js
```diff
@@ -0,0 +1,108 @@
...
+                  cache.put('style.css', response);
+                  // Inform any open pages that a new CSS file is available on
+                  // the server
+                  notifyPageOfCSSUpdate(true);
+                } else {
+                  notifyPageOfCSSUpdate(false);
+                }
+              });
+            } else {
+              notifyPageOfCSSUpdate(false);
```
delapuente: As this is by far less code and more simple, can we reverse the order of branches in the `if` checks?

update-css/sw.js
```diff
@@ -0,0 +1,108 @@
...
+function simulateStyleCSSNetworkFetch() {
+  return getCurrentCSSFilename().then(function(filename) {
+    return caches.open(cacheName).then(function(cache) {
+      // return the specified CSS file from the cache
+      return cache.match(filename);
+    });
+  });
+}
+
+function getCurrentCSSFilename() {
```
delapuente: If the _cache-as-db_ trick is an implementation detail I would hide that implementation detail. When solving #14, in #39, we restricted docco to the js of the first level, so you could use `importScripts(...)` and create something like `get().then()` and `put().then()` to actually hide these details (the we can open a followup to implement this with IndexedDB).

update-css/sw.js
delapuente: When checking the ETag (or any other _smart check_ as you say) I think it is very important to do this only **without completing** the request (if you have already completed the request, simply cache the new copy). As there is no cancellable fetchs (yet), it should be convenient to make a `HEAD` fetch. Does it make sense?

digitarald:
@TimAbraldes @delapuente what are the next steps here?

TimAbraldes:
> @TimAbraldes @delapuente what are the next steps here?

I picked this up today and am working on an updated PR. We may change the name/story associated with it but I think it'll make for a good recipe. I'll push some code tonight so that it's open for comment.

TimAbraldes:
As mentioned elsewhere, there's some confusion about what would satisfy the requirements for #9. This updated PR (not yet functional) is somewhat similar to stale-while-revalidate but it also incorporates a JSON file that lets the client know exactly what files need to be cached. This helps avoid partial updates (e.g. mismatched CSS files). I'm not sure what to call this recipe, or whether it will fulfill the requirements for #9... or whether this is a useful example for real web app builders... but it does seem to have some characteristics that aren't covered in other recipes.

TimAbraldes:
In the newest push, the functionality is basically working. I need to update many of the comments and the README, and I need to flesh out the recipe with more interesting example files (fonts, more css, etc). Also the "notify the page" functionality only seems to be working about half the time. Probably not worth reviewing at this point as there are many changes coming.

It turns out that partial updates are still possible: If the service worker finds updated CSS/fonts/etc., it will atomically update its cache, but the recipe doesn't yet protect already-loaded pages against fetching these new items, even though they won't match with the other files that they have already loaded. I'm devising a solution for this.

delapuente:
@TimAbraldes this is a real interesting recipe. Correct me if I'm wrong but you are trying to solve the problem of updating the application without passing through the service worker life-cycle. Isn't it? So this is why you recipe is similar to other as JSON Cache but it introduces a very important difference: the worker **never changes**. Just the update list changes.

When the worker never changes, you're decoupling the Application Update Process from Service Worker Update process (the [_roll your own_](https://github.com/mozilla/oghliner/pull/224#issuecomment-161736738) mode explained in that comment). They are now two different concepts. The latter only happens once while the second happens from time to time (in your demo, each time you click on _toggle version_ button).

This separated application life-cycle is, for instance, what [offliner](https://github.com/delapuente/offliner#updating) is trying to achieve and you will see a lot of similarities between both service workers, yours and mine.

So the use case your recipe is trying to solve is far beyond which kind of resources we are updating or caching, it is more about providing a separated update life-cycle for the Web Application:

> As a developer, I don't want to update the service worker each time I want to update my web application but simply instruct it to perform a resource update under certain conditions.

The solution can be articulated as:

> As you don't alter the service worker, it will never pass through installation nor activation again so you need to emulate these stages. Your client code will instruct the service worker to perform an update so it will pass through a custom install (by caching the resources again in a separated cache). After installing, notify the clients to say an activation is required. The client will instruct the service worker to perform the custom activation stage (removing the old cache and making the new one to be the active one). Finally notify the clients to inform the update is complete.  

If all this is correct (let me know if not) then the optimal solution is not trivial (offliner is a good try, xP) and it is beyond the scope of a single recipe but we can end with a sub-optimal demo solution. What do you think @TimAbraldes ?

TimAbraldes:
> @TimAbraldes this is a real interesting recipe. Correct me if I'm wrong but you are trying to solve the problem of updating the application without passing through the service worker life-cycle. Isn't it? So this is why you recipe is similar to other as JSON Cache but it introduces a very important difference: the worker **never changes**. Just the update list changes.

Yup! Though honestly in a real-world (non recipe) situation I would have preferred to update the Service Worker when updating the cached files. That's not possible in a recipe (we can't rewrite/modify the server version), though we could maybe emulate it by having two separate service worker JS files and choosing which one to register in the page.

> When the worker never changes, you're decoupling the Application Update Process from Service Worker Update process (the [_roll your own_](https://github.com/mozilla/oghliner/pull/224#issuecomment-161736738) mode explained in that comment). They are now two different concepts. The latter only happens once while the second happens from time to time (in your demo, each time you click on _toggle version_ button).

> This separated application life-cycle is, for instance, what [offliner](https://github.com/delapuente/offliner#updating) is trying to achieve and you will see a lot of similarities between both service workers, yours and mine.

Yeah, I wish I had read through the offliner code before writing this recipe!

> So the use case your recipe is trying to solve is far beyond which kind of resources we are updating or caching, it is more about providing a separated update life-cycle for the Web Application:

>> As a developer, I don't want to update the service worker each time I want to update my web application but simply instruct it to perform a resource update under certain conditions.

> The solution can be articulated as:

>> As you don't alter the service worker, it will never pass through installation nor activation again so you need to emulate these stages. Your client code will instruct the service worker to perform an update so it will pass through a custom install (by caching the resources again in a separated cache). After installing, notify the clients to say an activation is required. The client will instruct the service worker to perform the custom activation stage (removing the old cache and making the new one to be the active one). Finally notify the clients to inform the update is complete.  

Yup, this very accurately describes what the code in the recipe is doing.

> If all this is correct (let me know if not) 

You've definitely got it! Thanks for taking a thorough look and providing thoughtful feedback!

> then the optimal solution is not trivial (offliner is a good try, xP) and it is beyond the scope of a single recipe but we can end with a sub-optimal demo solution. What do you think @TimAbraldes ?

Yeah I completely agree. It's doubtful that, for a recipe, we'll be able to produce a solution that we consider "optimal." In a real-world scenario, the Service Worker would necessarily be a lot more complicated than the recipe version, but this can hopefully be instructive when someone is looking to add functionality to their Service Worker.

delapuente:
> That's not possible in a recipe (we can't rewrite/modify the server version), though we could maybe emulate it by having two separate service worker JS files and choosing which one to register in the page.

We actually can: in your recipe, you can add server side support by writing a `server.js` file. See [api-analytics](https://github.com/mozilla/serviceworker-cookbook/blob/master/api-analytics/server.js#L48) for an example with templates and [lod-balancer](https://github.com/mozilla/serviceworker-cookbook/tree/master/load-balancer) for an example with simple sessions.

Anyway, let's the Service Worker unchanged as this is relevant for the value proposition of the recipe. What we can change in the server is the JSON file (it can be generated dynamically in addition to the CSS file and we could randomly generate colors for a couple of items).

You can add a button in the UI saying, _Generate new version of resources_ and this will instruct the server to use a new version.

TimAbraldes:
> > That's not possible in a recipe (we can't rewrite/modify the server version), though we could maybe emulate it by having two separate service worker JS files and choosing which one to register in the page.

> We actually can: in your recipe, you can add server side support by writing a `server.js` file. See [api-analytics](https://github.com/mozilla/serviceworker-cookbook/blob/master/api-analytics/server.js#L48) for an example with templates and [lod-balancer](https://github.com/mozilla/serviceworker-cookbook/tree/master/load-balancer) for an example with simple sessions.

Whaaaaaaaa! That's awesome!

> Anyway, let's the Service Worker unchanged as this is relevant for the value proposition of the recipe. What we can change in the server is the JSON file (it can be generated dynamically in addition to the CSS file and we could randomly generate colors for a couple of items).

Sure, that sounds like a fun way to make the recipe more interesting! And also I get to create a server.js :)

> You can add a button in the UI saying, _Generate new version of resources_ and this will instruct the server to use a new version.

Yup, similar to the current version but with additional interesting changes to the resulting resources/CSS

TimAbraldes:
Issues:
  - In Firefox the page sometimes fails to load and the error console mentions a specific bug. I'm not sure if this can be fixed in the recipe or if we have to wait for the platform bug to be fixed.
  - The message from the SW informing client pages that an update is available never seems to be received. I'm not sure what is causing this, but it would be nice to get that working.

The other remaining items for this recipe are polish/visual items, so I think it's just about ready for review.

@delapuente would you like to review this recipe?

PR #66, opened 2015-11-14
Include polyfills for startsWith and endsWith

marco-c:
startsWith and endsWith are unavailable on Node 0.12.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,7 @@ git:
 language: node_js
 node_js:
 - '4'
+- '0.12'
 deploy:
   provider: heroku
   api_key:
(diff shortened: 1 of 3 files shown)


digitarald:
Would it be easier to just replace them with regexp (aka the solution to every problem ;) ), @brendandahl ?

We also should make sure we run tests with 1.2 to catch those use earlier.

brendandahl:
Why do we need to support v.12?

digitarald:
Make it easier for contributors. We don't really depend on 4.x features so it wouldn't make sense to require it, right?

darkwing:
R+, works for me!

PR #86, opened 2015-11-18
Add more detailed comments to the push-clients recipe

marco-c:


marco-c:
I think the behavior of the recipe was changed when we started using the new template. It's supposed to open the recipe's demo page.

PR #91, opened 2015-11-18
List recipes in order of difficulty

marco-c:
Depends on #89.


darkwing:
This is an awesome idea.  I think we need some sort of indicator before we merge this though.  What about something like this?  

notification

Maybe only on the first one of each difficulty?  I can code that up, just sounding out thoughts.

darkwing:
Actually we should probably just have "Beginner", "Intermediate", and "Advanced" headers and three separate lists.  That would probably be easiest

marco-c:
My idea was to redesign the list after doing this and after #92, to clearly separate recipes belonging to different categories.

marco-c:
So something like:
```
Category 1
Beginner Recipe 1
Beginner Recipe 2
Intermediate Recipe

Category 2
Beginner Recipe
Intermediate Recipe
Advanced Recipe
```

I don't know if we should also clearly indicate which ones are easy and which more difficult, because it could make the interface too cluttered.

darkwing:
I was thinking we'd do:

```
Beginner
- Recipe 1
- Recipe 2
...

Intermediate
- Recipe 1
- Recipe 2
...
```

digitarald:
the usual chili spiciness color coding might work here, green, yellow, red? I would rather see top level categories for use cases (offline vs push)

shidhincr:
+1

marco-c:
sidebar

darkwing:
R+, this looks good!

PR #93, opened 2015-11-18
Add tabzilla, minor design tweeks

darkwing:
Let me know if the fade-in of Tabzilla is "too much".  I do it because, in Chrome, setting the `:before` background flashes a dark "block" before the image is rendered.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -178,6 +178,12 @@ gulp.task('build:js', ['clean'], function() {
     .pipe(gulp.dest('./dist'));
 });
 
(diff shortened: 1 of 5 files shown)


marco-c:
For some reason the tab doesn't look very good on my screen.

marco-c:
tab

marco-c:
The `oghliner` one as well. I guess Firefox doesn't support HiDPI screens very well on Linux.

marco-c:
Looks good to me. I'll merge it as I don't mind the fade-in. We can always revert the fade-in in a follow-up.

PR #97, opened 2015-11-18
Allow Analytics example to work better within cookbook site iframe

darkwing:
This isn't optimal but I'm struggling for a better solution.  The "Analytics API" demo appears wrong because the table gets populated after page load.

I'm really hoping there's an easier, cleaner solution that I'm not seeing.

### DIFF
--- a/api-analytics/index.js
+++ b/api-analytics/index.js
@@ -45,6 +45,11 @@ function showQuotations(collection) {
     quote = collection[index];
     table.appendChild(getRowFor(quote));
   }
+
+  // Specifically for the cookbook site :(
+  if (window.parent !== window) {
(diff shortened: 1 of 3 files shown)


TimAbraldes:
> Height is more so the issue -- how would modifying `width` help?

Sorry, totally meant `height`. Updating comment.

I tried this out in a branch and it seems to work, see https://github.com/TimAbraldes/serviceworker-cookbook/commit/74af6447ec166c4e4dc459e1161c55e24b8c86d8

darkwing:
I implemented what you have here but the JSON Cache demo, for example, gets cut off.  The problem is that using `height: 100%` on the IFRAME is it's 100% of the current page, not the IFRAME's contents. :/

PR #105, opened 2015-11-19
Test/build on Travis CI with Node.js 0.12 as well; only deploy with Node.js 4

marco-c:
I'm too curious to try this again. Don't mind it, I'll take care of reverting if it makes Heroku fail.


marco-c:
It's working!

PR #107, opened 2015-11-19
Keep Service Worker alive while handling the 'pushsubscriptionchange' event

marco-c:


darkwing:
What's the best way to test this?

marco-c:
In theory a service worker could be terminated at any time by the browser, `event.waitUntil` makes sure its lifetime is extended until asynchronous operations are finished.

I don't know if there's an easy way to purposely make it fail.

darkwing:
R+!  Worked great!

PR #103, opened 2015-11-19
Add category field to recipes

marco-c:
Fixes #92.

Suggestions welcome on how to categorize the recipes where I put "?".

PR #109, opened 2015-11-19
List by category

### LINKED ISSUE #104: List recipes by category and then in order of difficulty
After #91 and #92, we should split the recipes in categories and order them by difficulty.

marco-c:
Fixes #104.

Depends on #103.

categories

### DIFF
--- a/src/tpl/layout.html
+++ b/src/tpl/layout.html
@@ -20,10 +20,9 @@
 
         {% set curCategory = "" %}
         {% for recipe in recipes %}
-
-        {% if recipe.category != curCategory %}
-        <li class="category">{{ recipe.category }}</li>
-        {% endif %}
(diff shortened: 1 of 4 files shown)


src/tpl/layout.html
```diff
@@ -17,7 +17,14 @@
 
       <ul class="recipes">
         <li><a href="index.html">Introduction</a></li>
+
+        {% set curCategory = "" %}
         {% for recipe in recipes %}
+
+        {% if recipe.category != curCategory %}
```
brendandahl: Nit: indention of 'if' and the '<li'>

darkwing:
R+ when nit fixed, looks awesome!

marco-c:
Fixed in 095d81b!

Issue #108, opened 2015-11-19
Wrap code (at 80 chars) in sample code?


awalGarg:
Heyo! Thanks for the really awesome cookbook :)

I was wondering if the code samples could be restricted to a specific line width, say 80 chars?

For example, https://serviceworke.rs/push-simple_index_doc.html introduces horizontal scrolling on my 1366x768 laptop screen.

This isn't a big issue at all, but *if* owners think it would be "okay" to have this, I'll be ready with a PR ;-)

<3, Awal

marco-c:
Yes, Awal. I think it would be great! Lines are too long even for my 1920x1080 screen :smile: 

It'd be great if you could also use the Q-style for promises, which makes lines shorter.
For example, this:
```JavaScript
return registration.pushManager.subscribe({ userVisibleOnly: true }).then(function(newSubscription) {
```
would become:
```JavaScript
return registration.pushManager.subscribe({ userVisibleOnly: true })
.then(function(newSubscription) {
```

mykmelez:
Even wrapping at 80 characters wouldn't avoid horizontal scrolling on my 15" MacBook Pro at 1440x900. On https://serviceworke.rs/push-simple_index_doc.html my browser displays only 69 characters of the code:

<img width="1440" alt="screen shot 2015-11-19 at 11 23 39" src="https://cloud.githubusercontent.com/assets/305455/11281683/73a857b6-8eb0-11e5-8bb3-32ce178fcbea.png">

(I can hide the sidebar, which works around the problem, but only until I reload the page.)

marco-c:
On my laptop, I can see 101 characters.

awalGarg:
@marco-c 

> It'd be great if you could also use the Q-style for promises, which makes lines shorter.

sure

@mykmelez 

> (I can hide the sidebar, which works around the problem, but only until I reload the page.)

hmm, we can use localstorage to remember state of sidebar across reloads? ;)
TBH, something below 80 chars sounds a wee bit too restrictive to me.

digitarald:
Lets add a lint rule for 80 characters then.

mykmelez:
> hmm, we can use localstorage to remember state of sidebar across reloads? ;)

That's a good idea!

> TBH, something below 80 chars sounds a wee bit too restrictive to me.

Indeed. Even 80 chars feels too limiting, even though well-formatted programs will rarely have lines that exceed that limit. It might be ok for a series of recipes, though.

[...]

PR #119, opened 2015-11-20
remember nav bar visibility using localstorage

awalGarg:
Initially discussed in https://github.com/mozilla/serviceworker-cookbook/issues/108#issuecomment-158184177, https://github.com/mozilla/serviceworker-cookbook/issues/108#issuecomment-158192392 and https://github.com/mozilla/serviceworker-cookbook/issues/108#issuecomment-158318447

### DIFF
--- a/src/js/layout.js
+++ b/src/js/layout.js
@@ -7,7 +7,9 @@ function resizeIframe(iframe) {
 (function() {
 
   document.getElementById('navToggle').addEventListener('click', function() {
-    document.querySelector('.book').classList.toggle('with-nav');
+    var book = document.querySelector('.book');
+    book.classList.toggle('with-nav');
+    localStorage.setItem('withNav', Number(book.classList.contains('with-nav')));
   });
 
   // This is a hacky way to highlight the current active tab. This should
@@ -25,6 +27,13 @@ function resizeIframe(iframe) {
   }
 
 })();
+
+document.addEventListener('DOMContentLoaded', function() {
+  document.querySelector('.book').classList.toggle(
+    'with-nav',
+    Number(localStorage.getItem('withNav'))
+  );
+});
 
 // Marking as loaded triggers tabzilla fade-in
 window.addEventListener('load', function() {
--- a/src/js/layout.js
+++ b/src/js/layout.js
@@ -8,8 +8,7 @@ function resizeIframe(iframe) {
 
   document.getElementById('navToggle').addEventListener('click', function() {
     var book = document.querySelector('.book');
-    book.classList.toggle('with-nav');
-    localStorage.setItem('withNav', Number(book.classList.contains('with-nav')));
+    localStorage.setItem('hideNav', 1-Number(book.classList.toggle('with-nav')));
   });
 
   // This is a hacky way to highlight the current active tab. This should
@@ -31,7 +30,7 @@ function resizeIframe(iframe) {
 document.addEventListener('DOMContentLoaded', function() {
   document.querySelector('.book').classList.toggle(
     'with-nav',
-    Number(localStorage.getItem('withNav'))
+    1-Number(localStorage.getItem('hideNav'))
   );
 });
 


src/js/layout.js
```diff
@@ -28,6 +30,8 @@ function resizeIframe(iframe) {
 
 window.addEventListener('load', function() {
   document.body.classList.add('loaded');
+  var navStatus = Number(localStorage.getItem('withNav'));
+  document.querySelector('.book').classList.toggle('with-nav', navStatus);
```
darkwing: We probably don't need `navStatus` as a var, or we could hoist it.

darkwing:
localStorage can be a pain in private/incognito windows -- should that be a concern?

awalGarg:
> localStorage can be a pain in private/incognito windows -- should that be a concern?

@darkwing no, that only happens if we do some convoluted operations on data returned by `.getItem` which wasn't previously set (by, say `.setItem`). Calling `Number` on `null` is perfectly valid and would simply fail to toggle the class.

awalGarg:
r = review? Sorry not very familiar with OSS contribution ^^

darkwing:
I like the idea, @awalGarg, but I think we should find a way to make the execution better.  The frequent slide on each link click is a bit jarring.  It may behoove us to make those top link clicks AJAX-powered so that the toggle doesn't happen.  Or we could add a session var to simply not show the sidebar on initial page load.  I'll continue to think on this.

awalGarg:
@darkwing I agree. Moving out the toggling from window.onload to document.onDOMContentLoaded seems to fix this, like so:

```
diff --git a/src/js/layout.js b/src/js/layout.js
index c19f912..372ff41 100644
--- a/src/js/layout.js
+++ b/src/js/layout.js
@@ -28,12 +28,19 @@ function resizeIframe(iframe) {
 
 })();
 
-window.addEventListener('load', function() {
-  document.body.classList.add('loaded');
-  document.querySelector('.book').classList.toggle(
+document.addEventListener('DOMContentLoaded', function() {
+  var book = document.querySelector('.book');
+  var nav = document.querySelector('nav');
+  nav.style.display = 'hidden';
+  book.classList.toggle(
     'with-nav',
     Number(localStorage.getItem('withNav'))
   );
+  nav.style.display = 'initial';
+});
+
+window.addEventListener('load', function() {
+  document.body.classList.add('loaded');
 });
 
 (function()  {
```

awalGarg:
err, ignore the .hidden stuff. I messed up some stuff.

awalGarg:
@darkwing I added the cleaned up domcontentloaded patch and rebased on latest master. Can you test now? The "jarring" isn't present anymore locally.

darkwing:
Wow @awalGarg , this is very close.  The only issue I spot is that the first viewing of the site has the sidebar hidden; we should really be showing the sidebar by default if the user hasn't been on the site yet.

awalGarg:
Hmm, that should be an easy fix but might make the code verbose. I'll see if there is a better way in a few hours.

darkwing:
What if we reverse the philosophy?  If we call it "hideNav" and set it to "1" to hide the nav, then the nav will display by default when the user hits the site the first time! :)

awalGarg:
@darkwing yeah that works indeed, good call :)

awalGarg:
Also just remembered that `.toggle` returns state anyways, so we don't need another call to `.contains`. Fixed that.

darkwing:
Awesome, this is much better!  Spending a tiny bit more time thinking on this; the "close" is un-noticeable in Chrome but a tad slower and noticeable in Firefox.  Going to see if there's anything more I can think of.  Great work @awalGarg , I'll have an update for you on this soon!

awalGarg:
@darkwing Hmm, I actually didn't test this in Firefox. My bad.

One way I see to solve this, is rather peculiar, and rather hacky. We know that script parsing is 100% blocking and the browser doesn't parse the page further until it completely executes current script (specially inline ones).

So, what we can do here, is move the `with-nav` class to the body from the `div.book` element. And then just before the `div.book` declaration, we can add the code which is responsible for toggling the class. Example patch:

```
diff --git a/src/css/style.css b/src/css/style.css
index f39035e..df3a07a 100644
--- a/src/css/style.css
+++ b/src/css/style.css
@@ -103,7 +103,7 @@ body {
 }
 
 @media (min-width: 600px) {
-  .book.with-nav main {
+  body.with-nav main {
     left: 300px;
   }
   .book main {
@@ -120,7 +120,7 @@ body {
     left: -100%;
     transition: left 250ms ease;
   }
-  .book.with-nav main {
+  body.with-nav main {
     transform: translate(calc(100% - 60px),0px);
   }
   .book main {
@@ -128,7 +128,7 @@ body {
   }
 }
 
-.book.with-nav nav {
+body.with-nav nav {
   left: 0;
 }
 
diff --git a/src/js/layout.js b/src/js/layout.js
index 2745003..f535e19 100644
--- a/src/js/layout.js
+++ b/src/js/layout.js
@@ -7,7 +7,7 @@ function resizeIframe(iframe) {
 (function() {
 
   document.getElementById('navToggle').addEventListener('click', function() {
-    document.querySelector('.book').classList.toggle('with-nav');
+    localStorage.setItem('hideNav', 1-Number(document.body.classList.toggle('with-nav')));
   });
 
   // This is a hacky way to highlight the current active tab. This should
diff --git a/src/tpl/layout.html b/src/tpl/layout.html
index bcb57d3..9f6cc15 100644
--- a/src/tpl/layout.html
+++ b/src/tpl/layout.html
@@ -9,8 +9,13 @@
   <link rel="stylesheet" href="bundle.css">
   <script type="text/javascript" src="bundle.js" defer></script>
 </head>
-<body>
-
+<body class="with-nav">
+<script type="text/javascript">
+  document.body.classList.toggle(
+    'with-nav',
+    1-Number(localStorage.getItem('hideNav'))
+  );
+</script>
   <script>
   (function(i,s,o,g,r,a,m){i['GoogleAnalyticsObject']=r;i[r]=i[r]||function(){
   (i[r].q=i[r].q||[]).push(arguments)},i[r].l=1*new Date();a=s.createElement(o),
@@ -21,7 +26,7 @@
   ga('send', 'pageview');
   </script>
 
-  <div class="book with-nav">
+  <div class="book">
     <nav><div class="recipes-cover">
       <div id="tabzilla">
         <a href="https://www.mozilla.org/">Mozilla</a>
```

I am not sure if this is the best solution, though.

awalGarg:
well, this patch is against my local copy of origin's master. dunno what commit it is at :/

darkwing:
We can give this a go and monitor going forward.

PR #120, opened 2015-11-20
wrap code to 80 chars and add linting rule

awalGarg:
Initially discussed in https://github.com/mozilla/serviceworker-cookbook/issues/108 (https://github.com/mozilla/serviceworker-cookbook/issues/108#issuecomment-158186148)

I'll push more commits as I wrap around more parts of the codebase. Shouldn't take too long.


awalGarg:
As of now, `gulp test` gives me 67 relevant max line warnings (1 is an alert warning, and some from `gulpfile.js` so I am not counting those).

Most of those are from comment sections, large urls in JS string literals etc. I have willingly left out the comment ones since docco will wrap the prose area automatically avoiding any horizontal scrolling issues, which is mostly the spirit of this PR.

I am not exactly sure what to do with declarations like https://github.com/mozilla/serviceworker-cookbook/blob/master/live-flowchart/service-worker-util.js#L28 though. Are the named functions really required? (pinging @franciov (from `git blame`)). With ES6 spec, functions are more rigorously assigned `name`s, and AFAIK, all those function expressions would automatically get their corresponding property names in any ES6 compliant JS engine. Just my 2 cents.

marco-c:
I think you can remove those names.

awalGarg:
ah well, I shall do that then

digitarald:
@marco-c can you merge please when this is r+?

.eslintrc
```diff
@@ -5,6 +5,7 @@
     "comma-dangle": 0,
     "func-names": 0,
     "vars-on-top": 0,
-    "no-use-before-define": 0
+    "no-use-before-define": 0,
+    "max-len": [1, 80, 4]
```
marco-c: Could you add ignoreComments and avoid wrapping comments?

api-analytics/index.js
```diff
@@ -13,12 +11,15 @@ document.getElementById('add-form').onsubmit = function() {
...
-  var quoteAuthor = document.getElementById('quote-author').value.trim() || 'Anonymous';
+  var quoteAuthor = document.getElementById('quote-author').value.trim()
+                    || 'Anonymous';
   var quote = { text: newQuote, author: quoteAuthor };
   var headers = { 'content-type': 'application/json' };
 
   // Send the API request. In this case, a `POST` on `quotations` collection.
-  fetch(ENDPOINT, { method: 'POST', body: JSON.stringify(quote), headers: headers })
+  fetch(ENDPOINT, {
+    method: 'POST', body: JSON.stringify(quote), headers: headers
```
marco-c: Nit: Since you need to wrap this, could you make it:
```
fetch(ENDPOINT, {
  method: 'POST',
  body: JSON.stringify(quote),
  headers: headers,
```

api-analytics/server.js
```diff
@@ -103,7 +105,9 @@ function summarizeLogs() {
   // Here we perform the aggregations.
   var aggregations = requestsLog.reduce(function(partialSummary, entry) {
     if (!(entry.url in partialSummary)) {
-      partialSummary[entry.url] = { url: entry.url, GET: 0, POST: 0, DELETE: 0 };
+      partialSummary[entry.url] = {
+        url: entry.url, GET: 0, POST: 0, DELETE: 0
```
marco-c: Nit: here like in the previous case.

cache-then-network/index.js
```diff
@@ -59,7 +59,9 @@ function enableUI() {
 }
 
 function updatePage(data) {
-  dataElement.textContent = 'User ' + data[0].actor.login + ' modified repo ' + data[0].repo.name;
+  var login = data[0].actor.login;
+  var name = data[0].repo.name;
+  dataElement.textContent = 'User ' + login + ' modified repo ' + name;
```
marco-c: Nit: you could do:
```

File: cache-then-network/index.js [hunk 5]
```diff
@@ -133,7 +135,9 @@ getDataButton.addEventListener('click', function handleClick() {
   // in Firefox or Chrome. See bug 1120715 for the Firefox
   // implementation status.
   var cacheBuster = Date.now();
-  var networkFetch = fetch(dataUrl + '?cacheBuster=' + cacheBuster, { mode: 'cors', cache: 'no-cache' }).then(function(res) {
+  var networkFetch = fetch(dataUrl + '?cacheBuster=' + cacheBuster, {
+    mode: 'cors', cache: 'no-cache'
```
marco-c: Nit: here like in the previous case.

gulpfile.js
```diff
@@ -141,10 +145,16 @@ gulp.task('build:intros', ['clean'], function() {
   return Promise.all(recipes.map(function(recipe) {
     return fsReadFile(recipe.slug + '/README.md', 'utf8')
     .then(function(readme) {
-      return renderFile('./src/tpl/intro.html', { markdown: marked(readme) });
+      return renderFile(
+        './src/tpl/intro.html', { markdown: marked(readme) }
+      );
```
marco-c: Nit:
```
      return renderFile('./src/tpl/intro.html', {
        markdown: marked(readme),
      });

File: live-flowchart/app.js [hunk 7]
```diff
@@ -68,38 +78,41 @@ App.prototype.enableCoolFeatures = function enableCoolFeatures() {
...
-      this.disableServiceWorkerRegistration // error
+      this.enableSWRegistration, // success
+      this.disableSWRegistration // error
   );
 };
 
 // <h3>Enable Service Worker Registration</h3>
 // Enable the possibility for the user to register a service worker
-App.prototype.enableServiceWorkerRegistration = function enableServiceWorkerRegistration() {
+App.prototype.enableSWRegistration = function enableSWRegistration() {
```
marco-c: You couly simply remove the name of the function

api-analytics/server.js
```diff
@@ -9,17 +9,18 @@ var swig = require('swig');
 // This is the collection of logs.
 var requestsLog = [];
 
-// And these are the default quotations. `makeDefaults()` simply add the id and the
+// And these are the default quotations.
```
marco-c: No need to wrap comment

marco-c: Here and in other places

awalGarg: ah right, I will fix those in a moment, thanks

marco-c: No, thank you! This must be painful work :smile:

gulpfile.js
```diff
@@ -141,10 +145,16 @@ gulp.task('build:intros', ['clean'], function() {
...
     return fsReadFile(recipe.slug + '/README.md', 'utf8')
     .then(function(readme) {
-      return renderFile('./src/tpl/intro.html', { markdown: marked(readme) });
+      return renderFile('./src/tpl/intro.html', {
+        markdown: marked(readme)
+      });
     })
     .then(function(content) {
-      return template.writeFile('./dist/' + recipe.slug + '.html', content, { currentRecipe: recipe });
+      return template.writeFile(
```
marco-c:
```
      return template.writeFile('./dist/' + recipe.slug + '.html', content, {
        currentRecipe: recipe
      });
```

marco-c: Here and the next.

marco-c: If it becomes more than 80, please ignore my comment :)

live-flowchart/app.js
```diff
@@ -16,27 +16,33 @@ App.prototype.constructor = function constructor() {
   this.swUtil = new SWUtil();
 
   // register click events
-  document.querySelector('#reloadapp').addEventListener('click', function reloadappHandler() {
-    window.location.reload();
-  });
+  document.querySelector('#reloadapp').addEventListener('click',
+    function reloadappHandler() {
```
marco-c: Would it fit in one line if you removed the name here?

live-flowchart/app.js
```diff
@@ -16,27 +16,33 @@ App.prototype.constructor = function constructor() {
...
   if (this.swUtil.areServiceWorkersSupported()) {
-    document.querySelector('#swinstall').addEventListener('click', function swinstallHandler() {
-      thisapp.enableCoolFeatures();
-    });
-
-    document.querySelector('#swuninstall').addEventListener('click', function swuninstallHandler() {
-      thisapp.disableCoolFeatures();
-    });
+    document.querySelector('#swinstall').addEventListener('click',
+      function swinstallHandler() {
```
marco-c: Same here

live-flowchart/app.js
```diff
@@ -16,27 +16,33 @@ App.prototype.constructor = function constructor() {
...
+        thisapp.enableCoolFeatures();
+      }
+    );
+
+    document.querySelector('#swuninstall').addEventListener('click',
+      function swuninstallHandler() {
```
marco-c: And here

api-analytics/server.js
awalGarg: hah, I just did a soft reset to the last master commit, and started staging interactively :)

gulpfile.js
awalGarg: It closed at 78 on the first one, but not on the next one, so left the next one as is.

live-flowchart/app.js
awalGarg: It does, done

live-flowchart/app.js
awalGarg: it doesn't here :/ left as is

awalGarg:
@marco-c can you recheck? I just pushed the updated commits

marco-c:
Thank you! I've merged it manually because there were some conflicts.
I've only reverted the changes to parseRecipes.js and gulpfile.js because we don't need to enforce this rule for those files.

PR #121, opened 2015-11-20
Don't build on Heroku; build on Travis and deploy the result

marco-c:


marco-c:
This should make deployment to Heroku faster and should waste less resources.

digitarald:
Great, @marco-c we should do the same for platatus

r? @brendandahl @darkwing

digitarald:
Downside I see that the build on travis might be a development build at one point and heroku would get a production build. For sw.rs we don't have that config yet.

marco-c:
> Downside I see that the build on travis might be a development build at one point and heroku would get a production build. For sw.rs we don't have that config yet.

If that happens, we could configure Travis to do:
1) Build development build
2) Run tests
3) Clean
4) Build production build
5) Deploy to Heroku

It'd still be faster because Heroku wouldn't need to install any devDependency.

PR #122, opened 2015-11-20
Resolve #69 - Add 'sync-online' recipe

### LINKED ISSUE #69: Recipe: Sync when online
Let the user do some work on a resource when offline and upload the changes when online.

darkwing:
Getting a stupid JSON parsing error from the initial fetch and I have no clue why.  Work in progress.


sync-online/index.js
```diff
@@ -0,0 +1,68 @@
...
+  });
+} else {
+  init();
+}
+
+// Function to get content for page
+function init() {
+  return fetch('?get')
+    .then(function(response) {
+      return response.json();
```
darkwing: Seeing this error:

```
// Error:
// Uncaught (in promise) SyntaxError: Unexpected token o
// Because the fetch is seeing "[object Object]""
```

But I'm sending back:

```
return new Response({
            ok: true,
            status: '200',
            statusText: 'OK',
            headers: { 'content-type': 'application/json' },
            body: JSON.stringify(content)
          });
```
... so I'm confused.

darkwing: I'm wondering if it's responding with `managePostContent(null)`'s promise now...

darkwing:
ToDo:

[ ] -  Lint
[ ] -  Add indicator so user knows what's going on (when saving is happening)
[ ] - Think of a way to manage who to trust as having newer content:  the server or indexeddb?  (probably check last save time)

sync-online/index.js
```diff
@@ -0,0 +1,66 @@
+// Our textarea
+var textarea = document.querySelector('textarea');
+
+// Register the ServiceWorker
```
darkwing: I know all of this `init()` stuff looks odd but I *really* don't want the fetch to happen before the service worker has had a chance to work.

sync-online/service-worker.js
```diff
@@ -0,0 +1,147 @@
+// [Working example](/serviceworker-cookbook/sync-online/).
+
+var CACHE_NAME = 'dependencies-cache';
+
+// Files required to make this app work offline
+var REQUIRED_FILES = [
+  '/',
+  '/index.html',
```
darkwing: Can remove this once https://github.com/mozilla/serviceworker-cookbook/pull/126 is merged.

delapuente:
Hey @darkwing I implemented this one #138 but I'm making some changes to differentiate your use case from my use case as I see them as two solutions for the same problem, yours in an specific way, mine dealing with delaying several requests. I really like your implementation but I'm leaving you some comments. :)

sync-online/service-worker.js
```diff
@@ -0,0 +1,147 @@
...
+
+self.addEventListener('activate', function(event) {
+  console.log('[activate] Activating ServiceWorker!');
+
+  // Calling claim() to force a "controllerchange" event on navigator.serviceWorker
+  console.log('[activate] Claiming this ServiceWorker!');
+  event.waitUntil(self.clients.claim());
+});
+
+function getDatabase() {
```
delapuente: You can use [localForage](https://github.com/mozilla/localForage) instead hiding all the database logic.

sync-online/service-worker.js
```diff
@@ -0,0 +1,147 @@
...
+        })
+      );
+    }
+  } else {
+    // Make network request
+    event.respondWith(fetch(event.request));
+  }
+});
+
+self.addEventListener('activate', function(event) {
```
delapuente: I would tend to group `install` and `activate` code and keep them together with only one explanation for them.

sync-online/index.js
```diff
@@ -0,0 +1,66 @@
...
+    console.error('The service worker could not be registered');
+    init();
+  });
+} else {
+  init();
+}
+
+// Function to get content for page
+function init() {
+  return fetch('?get')
```
delapuente: You could use a post to a fake API such as `GET api/contents` for retrieving the content...

sync-online/index.js
```diff
@@ -0,0 +1,66 @@
...
+      return response.json();
+    })
+    .then(function(obj) {
+      textarea.value = obj.content;
+
+      // Now that the content is populated and we know the "setup" is good,
+      // we can set off autosaving every 5 seconds
+      setInterval(function() {
+        // Thanks to our service worker, "saving" will be as simple fetch!
+        var request = {
```
delapuente: And a POST to a fake API such as `POST api/contents` with the body you're providing to do this.

sync-online/service-worker.js
```diff
@@ -0,0 +1,145 @@
...
+  if (self.indexedDB && event.request.url.indexOf('?') !== -1) {
+    // Save!
+    if (event.request.method === 'POST') {
+      // Send the fetch to the server if online
+      // We aren't concerned about the state of this because we have IndexDB as a backup
+      fetch(event.request.clone());
+
+      event.respondWith(
+        event.request.json().then(function(json) {
+          return managePostContent(json.content).then(function(content) {
```
darkwing: Can someone shed a bit of light on why this POST is 404'ing on and off?  I find it odd that it's not consistent.

sync-online/service-worker.js
```diff
@@ -0,0 +1,145 @@
+// [Working example](/serviceworker-cookbook/sync-online/).
+
+var CACHE_NAME = 'dependencies-cache';
+
+// Files required to make this app work offline
+var REQUIRED_FILES = [
+  '/',
```
darkwing: Since I'm using a scope of '.' for the SW registration, shouldn't '/' == '/sync-online/'

darkwing:
I've rebased this and added a few questions; insights from @marco-c or @delapuente would be awesome.  Thanks!

sync-online/index.js
```diff
@@ -0,0 +1,79 @@
...
+    .then(function(response) {
+      return response.json();
+    })
+    .then(function(obj) {
+      textarea.value = obj.content;
+
+      // Now that the content is populated and we know the "setup" is good,
+      // we can set off autosaving every 5 seconds
+      //setInterval(function() {
+      setTimeout(function() {
```
darkwing: Temporarily switched to `setTimeout` for the sake of debugging

sync-online/service-worker.js
darkwing: Avoided that to teach myself indexedDB.  Bit selfish, I can update that.

sync-online/index.js
marco-c: I think there's no need for all the extra checks here, fetch events are delayed if the service worker is activating.

sync-online/service-worker.js
marco-c: I think the scope only defines which URLs are controlled by the service worker. The URLs of the requests should be relative to the service worker script URL.

'/' is '/sync-online/' for both.

sync-online/service-worker.js
marco-c: Maybe the managePostContent promise is being rejected? You could add a rejection handler here and log the error

sync-online/index.js
delapuente: Yes, it's simple to use the `.ready` promise.

darkwing:
I'm going to close this, if only temporarily.  I'm not thrilled with my implementation and it may be best to start over :(  I'll think on this.

cheako:
@darkwing Thanks for this back in 2016.  Though today it's interfering with my searches for examples using https://developers.google.com/web/updates/2015/12/background-sync

Also I see that IndexedDB is accessed from inside the SW, I would advise against this because it prevents upgrading the DB...  That would lock the 'current' SW out of the database until the new version of the SW(that supported the new schema) is loaded and that's not something I believe one can do from inside the SW...  Only a window can reload it's self(thought one should warn the user first).

PR #128, opened 2015-11-24
Move template to more descriptive directory, register serviceworker in index.js

darkwing:
Having a "_template" directory isn't very helpful, so I've renamed it and called it out in the README


marco-c:
Need to update https://github.com/mozilla/serviceworker-cookbook/blob/master/gulpfile.js#L18

darkwing:
Good catch!  Updated!

PR #135, opened 2015-11-25
Call claim, don't bind

marco-c:


marco-c:
R? @delapuente

delapuente:
You're right.

PR #138, opened 2015-11-25
Request sync

### LINKED ISSUE #19: Recipe: request echoer
When loosing network connection, requests to the server can be cached keeping the order answering with place-holders responses. Later, when regaining connectivity, the service worker can replicate the server session .

delapuente:
Fix #19 
@marco-c do you mind to take a look at this recipe? If you have any doubt, ping me!


session-deferrer/README.md
```diff
@@ -0,0 +1,16 @@
+# REST API Session Deferrer
```
marco-c: Suggestion: requests deferrer instead of session deferrer?

session-deferrer/README.md
```diff
@@ -0,0 +1,16 @@
...
+This recipe shows how to record requests to a remote REST API while offline and defer them until regaining connectivity.
+
+## Difficulty
+Advanced
+
+## Use case
+As a web app developer, I want to provide functionality during offline, synchronizing with the server once the network is available again.
+
+## Solution
+Use a Service Worker to intercept requests. While offline, record the succesive requests in a queue to preserve the order and answer with fake responses. If online, flush the queue to replay the session and sync with the server.
```
marco-c: Nit: successive

session-deferrer/index.js
```diff
@@ -0,0 +1,158 @@
...
+  };
+  navigator.serviceWorker.register('service-worker.js');
+}
+
+// This forces a sync with the server. The POST request will never
+// reach the network. It is intended to be intercepted by the
+// Service Worker forcing it to do a sync with the actual remote
+// server.
+window.addEventListener('online', function() {
+  fetch('api/sync', { method: 'POST' }).then(function() {
```
marco-c: We should pick a style and always use the same for all recipes (in a followup).

session-deferrer/index.js
```diff
@@ -0,0 +1,158 @@
...
+        return quote;
+      }
+      return response.json();
+    })
+    .then(function(addedQuote) {
+      document.getElementById('quotations').appendChild(getRowFor(addedQuote));
+    });
+};
+
+// A simply `GET` (default operation for `fetch()`) is enough to retrieve
```
marco-c: Nit: simple

request-deferrer/README.md
```diff
@@ -0,0 +1,18 @@
+# Request Deferrer
+The recipe show how to enqueue requests while in offline in an _outbox-like_ buffer to perform the operations once the connection is regained.
```
marco-c: nit: "This recipe shows"

request-deferrer/index.js
```diff
@@ -0,0 +1,158 @@
...
+    };
+  };
+  navigator.serviceWorker.register('service-worker.js');
+}
+
+// This forces a sync with the server. The POST request will never
+// reach the network. It is intended to be intercepted by the
+// Service Worker forcing it to do a sync with the actual remote
+// server.
+window.addEventListener('online', function() {
```
marco-c: We should double-check that this works on all OSes.
See for example:
https://bugzilla.mozilla.org/show_bug.cgi?id=1228313
https://bugzilla.mozilla.org/show_bug.cgi?id=1228309

request-deferrer/index.js
```diff
@@ -0,0 +1,158 @@
...
+      return response.json();
+    })
+    .then(showQuotations);
+}
+
+// Fill the table with the quotations.
+function showQuotations(collection) {
+  var table = document.getElementById('quotations');
+  table.innerHTML = '';
+  for (var index = 0, max = collection.length, quote; index < max; index++) {
```
marco-c: Nit: forEach

request-deferrer/service-worker.js
```diff
@@ -0,0 +1,182 @@
...
+// So here is the idea. We will check if we are online or not. In case
+// we are not online, enqueue the request and provide a fake response.
+// Else, flush the queue and let the new request to reach the network.
+
+// This function factory does exactly that.
+function tryOrFallback(fakeResponse) {
+  // Return a handler that...
+  return function(req, res) {
+    // If offline, enqueue and answer with the fake response.
+    if (!navigator.onLine) {
```
marco-c: This could have the same problems of the `online` event.

request-deferrer/service-worker.js
```diff
@@ -0,0 +1,182 @@
...
+})));
+
+// Creation is another story. We can not reach the server so we can not
+// get the id for the new quotations. No problem, just say we accept the
+// creation and we will process it later, as soon as we recover connectivity.
+worker.post(root + 'api/quotations?*', tryOrFallback(new Response(null, {
+  status: 202
+})));
+
+// Lastly, this a virtual route to force the a sync. It is originated
```
marco-c: this is

request-deferrer/service-worker.js
```diff
@@ -0,0 +1,182 @@
...
+// by the client who is listening for the `online` event. This request
+// never reaches the network. It is here to be consistent but it is pure
+// virtual.
+worker.post(root + 'api/sync', function() {
```
marco-c: Why is this needed?

request-deferrer/service-worker.js
```diff
@@ -0,0 +1,182 @@
...
+  return flushQueue().then(function() {
+    return new Response();
+  });
+});
+
+// Start the service worker.
+worker.init();
+
+// By using Mozilla's localforage db wrapper, we count with
+// a fast setup for a versatile key, value database. We use
```
marco-c: Nit: key-value
Nit2: I think "count with" should be "can count on"

request-deferrer/service-worker.js
```diff
@@ -0,0 +1,182 @@
...
+// it to store queue of deferred requests.
+
+// Enqueue consists into add to the list a request. Due to the
```
marco-c: "consists of adding a request to the list"

request-deferrer/service-worker.js
```diff
@@ -0,0 +1,182 @@
...
+// limitations of IndexedDB, Request and Response objects can not
+// be saved so we need an alternative representations. This is
```
marco-c: They could be stored in a cache to avoid the complication of `serialize`

marco-c: (Just throwing the idea, you don't have to do it if you don't want to)

request-deferrer/index.html
```diff
@@ -0,0 +1,33 @@
...
+      font-weight: bold;
+    }
+  </style>
+</head>
+<body>
+  <h1>How to</h2>
+  <p>Try to add and delete some quotations.</p>
+  <p>Then go offline by disconnecting from Internet</p>
+  <p>And try to continue adding some quotes.</p>
+  <p>You see they have not a delete button because they are in queue</p>
```
marco-c: "You can see that they don't have a delete button"

request-deferrer/service-worker.js
```diff
@@ -0,0 +1,182 @@
...
+  var sending = Promise.resolve();
+  requests.forEach(function(serialized) {
+    console.log('Sending', serialized.method, serialized.url);
+    sending = sending.then(function() {
+      return deserialize(serialized).then(function(request) {
+        fetch(request);
+      });
+    });
+  });
+  return sending;
```
marco-c: Nit: it would be nice to use `reduce` here:
```JavaScript
return requests.reduce(function(prevPromise, serialized) {
    console.log('Sending', serialized.method, serialized.url);
    return prevPromise.then(function() {
      return deserialize(serialized).then(function(request) {
        fetch(request);
      });
    });
}, Promise.resolve());
```

P.S.: To send the requests in order, we should also wait the fetch promise to be resolved, so instead of `fetch(request)` -> `return fetch(request);`

request-deferrer/service-worker.js
delapuente: You're right, [per spec](http://www.w3.org/TR/service-workers/#cache-keys), keys are retrieved in _key insertion order_ so I will use a cache and simplify this point. Thank you!

request-deferrer/service-worker.js
delapuente: Yes, but `reduce()` tend to be more difficult to understand than simple loops. You're right with the `return fetch()`, nice catch.

request-deferrer/index.js
delapuente: I'll check them again. But I did on ArchLinux, and I did not see a problem. Such weird bug if it differs from one to another distribution.

marco-c: On Fedora, the `online` event is triggered when you enable WiFi, even before connecting (so in general it could be triggered even if you can't connect to any network).

request-deferrer/service-worker.js
delapuente: To force the synchronization when connectivity comes back. We post to sync from the client as the result [of `online` event](https://github.com/mozilla/serviceworker-cookbook/pull/138/files#diff-a04fecda16b7cbaab1fe88e4a97ea39bR22).

marco-c: Yes, I mean, why do we need this synchronization? Couldn't the client directly send the GET request to load the quotations as soon as he is online?

delapuente: You bet. At the beginning, it was not reloading but you're totally right, let me double-check it's not needed.

request-deferrer/service-worker.js
```diff
@@ -0,0 +1,183 @@
...
+});
+
+// Start the service worker.
+worker.init();
+
+// By using Mozilla's localforage db wrapper, we can count on
+// a fast setup for a versatile key-value database. We use
+// it to store queue of deferred requests.
+
+// Enqueue consists on adding a request to the list. Due to the
```
marco-c: consists of

request-deferrer/service-worker.js
```diff
@@ -0,0 +1,183 @@
...
+      queue = queue || [];
+      queue.push(serialized);
+      return localforage.setItem('queue', queue).then(function() {
+        console.log(serialized.method, serialized.url, 'enqueued!');
+      });
+    });
+  });
+}
+
+// Flush is a little more complicated. It consists into get
```
marco-c: consists of

request-deferrer/service-worker.js
```diff
@@ -0,0 +1,183 @@
...
+  return sending;
+}
+
+// Serialize is a little bit convolved due to headers is
+// a [map](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map)
+// and not a simple object.
+function serialize(request) {
+  var headers = {};
+  // `for(... of ...)` is ES6 notation but current browsers supporting SW, support this
+  // notation as well and this is the only way of retrieving all the headers.
```
marco-c: Nit: I think you could also use forEach here

delapuente:
I'm going to not implement the queue as a cache as currently, Cache API only works for `GET` requests.
Some discussion: https://github.com/slightlyoff/ServiceWorker/issues/693

request-deferrer/index.js
marco-c: Filed #151 as a follow-up.

request-deferrer/service-worker.js
marco-c: Talked with @delapuente on IRC, `forEach` isn't currently available for Headers.
Headers is not a Map though, so the comment needs to be fixed.

PR #140, opened 2015-11-26
Warn that message-relay recipe doesn't work yet in Chrome

marco-c:


message-relay/service-worker.js
```diff
@@ -1,23 +1,26 @@
...
-  self.clients.matchAll().then(function(clientList) {
-    clientList.forEach(function(client) {
-      // Skip sending the message to the client that sent it.
-      if (client.id === event.source.id) {
-        return;
-      }
-      client.postMessage({
-        client: event.source.id,
-        message: event.data
+  event.waitUntil(
```
darkwing: Each `letter` triggers a `TypeError: event.waitUntil is not a function` error here.

marco-c: In Chrome?

darkwing: Firefox Dev Edition:  
screenshot at nov 30 09-55-22

marco-c: OK, that explains it, since I'm using Nightly.

The correct thing to do is to use *event.waitUntil*, what should we do?
- Not use it;
- Use it only if it's defined;
- Use it anyway.

I'd do option 3 (because I think the recipes need to be as correct as possible) or option 2.

darkwing: I agree, but since we're adding a message for Chrome in this, should we also add a note for Firefox/Dev/Nightly?  This recipe looks broken otherwise.

marco-c: What about option 2 with a warning comment? This way it works in both nightly and dev edition, but the readers clearly see what's the correct behavior.

marco-c: (Unless the changeset that fixed this issue will be backported to dev edition, in that case we don't need to do anything special)

marco-c: https://bugzilla.mozilla.org/show_bug.cgi?id=1207068

marco-c: Looks like it's going to be uplifted to Aurora!

digitarald:
@marco-c can we show how to best feature detect the missing support and warn in the UI without falling over?

marco-c:
On Chrome basically we can't do:
```
  // Skip sending the message to the client that sent it.
  if (client.id === event.source.id) {
    return;
  }
```
So we can't detect if the message came from the client that sent it and we're going to send it to all clients.

marco-c:
Is the different behavior OK?
I could change the warning to say:
`Warning: on Chrome there's no way to detect which service worker client sent a message, so you're going to see an echo (the sender will receive the messages it sends, even if it isn't supposed to).`

digitarald:
> On Chrome basically we can't do:

Sounds like we can detect if the property got send correctly and if not log a warning?

marco-c:
I've added some more warnings in the inline documentation; added a console.log warning; avoided using event.waitUntil if not defined (in Chrome it isn't).

PR #148, opened 2015-11-29
Make push-client recipe open the demo page when the user clicks on a notification

marco-c:
Fixes regression from the cookbook page restructuring.

### DIFF
--- a/push-clients/service-worker.js
+++ b/push-clients/service-worker.js
@@ -53,7 +53,7 @@ self.addEventListener('notificationclick', function(event) {
       }
 
       // Otherwise, open a new page.
-      return self.clients.openWindow('./index.html');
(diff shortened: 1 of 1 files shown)


darkwing:
Should we add more logic here to check if the user was at `index.html` (they went directly to the demo) or `../push-clients_demo.html`?  Ideally we could make the demo work in both scenarios.

marco-c:
Instead of making the URL depend on a stored information (that looks too specific to the cookbook website), we could make it depend on some information passed by the server.
I think it would be better because I can see a real use case behind it (e.g. a newspaper website that sends a push notification with the URL of the news).

It would still introduce some complexity in the recipe.

digitarald:
I am for the best guess for now.

PR #162, opened 2015-12-04
Show curl command to send a notification in the push-simple recipe

marco-c:


push-simple/index.html
```diff
@@ -22,7 +26,11 @@
 Notification Time-To-Live: <input id='notification-ttl' type='number' value='0'></input> seconds
 </form>
 
-<button id="doIt">Try to conquer Italy!</button>
+<p>Either cick this button to ask the server to send a notification: <button id="doIt">Try to conquer Italy!</button></p>
```
darkwing: "click"

darkwing:
I received the following when attempting to curl:

https://gist.github.com/darkwing/3aaa444b4015cc534061

Does that relate to the following which is seen when starting the server: "If you want Chrome to work, you need to set the GCM_API_KEY environment variable to your GCM API key." ?

marco-c:
Yes, it only works with services that implement the Web Push standard (GCM doesn't), I've mentioned it in the documentation.

darkwing:
R+

PR #163, opened 2015-12-07
README cleanup for consistency

darkwing:
Using sentence case for consistency and other minor updates.

### DIFF
--- a/cache-then-network/README.md
+++ b/cache-then-network/README.md
@@ -5,10 +5,10 @@ This recipe illustrates methods to return network requests from either the cache
 ## Difficulty
 Beginner
 
-## Use case
+## Use Case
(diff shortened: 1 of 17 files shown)


cache-then-network/README.md
```diff
@@ -5,13 +5,13 @@ This recipe illustrates methods to return network requests from either the cache
...
 
-## Features and usage
+## Features and Usage
 
 - Configure form to allow or disallow cache and/or network
 - Click "Get new data" to execute requests
 
 ## Category
-Performance
+General Usage
```
marco-c: I'd keep "Performance" here, we might want to add more recipes in that category.

darkwing:
Updated!  I removed "Performance" because it was looking lonely with only one item but I guess that's motivation to create more, right?  :)

PR #167, opened 2015-12-08
Resolve #157 - Create sort order for sidebar

### LINKED ISSUE #157: Order categories
I think we should order them differently:
- General usage
- Offline
- More than offline
- Web Push

...Instead of:
- General usage
- More than offline
- Offline
- Web Push

darkwing:
I also changed "More than Offline" to "Beyond Offline"

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -22,18 +22,7 @@ var recipeSlugs = glob.sync('./!(dist|node_modules|src|_recipe_template)/').map(
 var srcRecipes = recipeSlugs.map(function makePath(name) {
   return './' + name + '/**';
 });
-
-var recipes = parseRecipes(recipeSlugs).sort(function(recipeA, recipeB) {
-  if (recipeA.category === recipeB.category) {
-    return recipeA.difficulty < recipeB.difficulty ? -1 : 1;
-  }
-
-  if (recipeA.category < recipeB.category) {
-    return -1;
-  }
-
-  return 1;
-});
+var recipes = parseRecipes(recipeSlugs);
 
 var template = (function() {
   function renderContent(content, options) {
--- a/parseRecipes.js
+++ b/parseRecipes.js
@@ -3,8 +3,16 @@ var glob = require('glob');
 var fs = require('fs');
 var assert = require('assert');
 
-module.exports = function(recipeSlugs) {
-  return recipeSlugs.map(function(recipe) {
+var difficulties = {
+  'Advanced': 3,
+  'Intermediate': 2,
+  'Beginner': 1
+};
+
+var categoryOrder = ['General Usage', 'Offline', 'Beyond Offline', 'Performance', 'Web Push'];
+
+module.exports = function(recipeSlugs) { 
+  var recipes = recipeSlugs.map(function(recipe) {
     var tokens = marked.lexer(fs.readFileSync(recipe + '/README.md', 'utf8'));
     assert(tokens.length > 1, recipe + ': README.md must have contents.');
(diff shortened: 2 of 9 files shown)


gulpfile.js
```diff
@@ -23,6 +23,8 @@ var srcRecipes = recipeSlugs.map(function makePath(name) {
   return './' + name + '/**';
 });
 
+var categoryOrder = ['General Usage', 'Offline', 'Beyond Offline', 'Performance', 'Web Push'];
```
marco-c: Could you add an assertion to check that a recipe belongs to one of these categories?

darkwing:
Updated!  Let me know if it's to your liking!

parseRecipes.js
```diff
@@ -27,13 +35,10 @@ module.exports = function(recipeSlugs) {
...
-      difficulty = 3;
-    } else if (difficultyTerm === 'Intermediate') {
-      difficulty = 2;
-    } else if (difficultyTerm === 'Beginner') {
-      difficulty = 1;
-    } else {
+    if(difficultyTerm in difficulties) {
+      difficulty = difficulties[difficultyTerm];
+    }
+    else {
```
marco-c: Huge nit: } else {

marco-c:
> Updated! Let me know if it's to your liking!

I like that you've moved everything in the parseRecipes function! You haven't added the assertion though :smiley:

darkwing:
Doh, had un-pushed commit.  Updated!

parseRecipes.js
```diff
@@ -58,5 +62,19 @@ module.exports = function(recipeSlugs) {
...
+    }
+
+    if (recipeA.category < recipeB.category) {
+      return -1;
+    }
+
+    return 1;
+  }).sort(function(recipeA, recipeB) {
+    // Sort recipes by preferred category
+    assert(categoryOrder.indexOf(recipeA.category) !== -1, recipeA.name + ': Unexpected category value "' + recipeA.category + '"');
```
marco-c: With `sort` some elements of the array could be missed. E.g. if you run this code:
```
[3, 4, 5].sort(function(a, b) {
  console.log(a);
  return a - b;
}
```
only *3* and *4* will be logged.

`forEach` would definitely be called for all the elements.

darkwing: Wow, I didn't know that; how can we use `forEach` to do the sort?

marco-c: You can use `sort` to sort and then `forEach` to assert (or the other way around).

darkwing: Ohh, your concern is the assert and not the correct sort?  OK, that makes sense.  Will update.

darkwing:
Grrr, the duplicate sort here either orders the categories wrong or the difficulties wrong, based on the order of the sort calls; this will require more changes.

marco-c:
Nit: I was thinking that instead of adding the `assert` with the `forEach`, you could add it in the mapping function (`recipeSlugs.map(function(recipe) {`), where other assertions are.

To fix the sorting problem, you can put everything in a single sorting function, something like:
```JavaScript
function(recipeA, recipeB) {
  // Sort recipes by preferred category
  if (categoryOrder.indexOf(recipeA.category) === categoryOrder.indexOf(recipeB.category)) {
    // Sort by difficulty
  } else if (categoryOrder.indexOf(recipeA.category) > categoryOrder.indexOf(recipeB.category)) {
    return 1;
  } else {
    return -1;
  }
}
```

darkwing:
Ahh, very nice.  Updated!

PR #171, opened 2015-12-11
Possible alternate design

darkwing:
Less green-on-green crime.

screenshot at dec 11 11-38-00

### DIFF
--- a/src/css/style.css
+++ b/src/css/style.css
@@ -98,7 +98,7 @@ body {
   z-index: 1;
   width: 300px;
   color: #364149;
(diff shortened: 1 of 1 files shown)


marco-c:
The yellow on black is a little annoying for my eyes.

fbender:
I like it though I agree the bright-yellow-on-black does not fit well. May you upload this as an alternate style so we can interate on it on-line?

darkwing:
Updated!  
screenshot at dec 14 10-02-18

PR #173, opened 2015-12-11
Add 'cache for performance' basic demo

darkwing:
Thought this was a nice, basic example of risk-free caching of assets for performance purposes.


cache-for-performance/service-worker.js
```diff
@@ -0,0 +1,74 @@
...
+self.addEventListener('install', function(event) {
+  // Perform install step:  loading each required file into cache
+  event.waitUntil(
+    caches.open(CACHE_NAME)
+      .then(function(cache) {
+        // Add non-essential files to cache if possible
+        // Since this doesn't "return", if any item fails, the
+        // service worker still installs properly with
+        // essential offline files
+        cache.addAll(PERFORMANCE_FILES);
```
darkwing: The main point of this demo is using addAll but not returning it, so it cannot interfere with required core files.

marco-c: I've been thinking about this and I'm not totally convinced that it is a good idea. Let's talk about it later.

darkwing:
Any verdict on this?  Shall we merge or close?,

marco-c:
My doubt was that this could make performance unpredictable and, if there's a failure while caching the "performance files", there's no way to recover it.

At the same time this recipe is meant to be simple, so I don't know if those problems actually matter.

darkwing:
I'm going to make the executive decision and close this; we shouldn't add anything we aren't sold on.  Thanks @marco-c !

PR #174, opened 2015-12-11
Move render-store recipe to the Performance category

marco-c:


delapuente:
Nice and now we don't need to remove that category.

Issue #184, opened 2015-12-27
help my code reading!!!


lv7777:
https://serviceworke.rs/api-analytics_service-worker_doc.html

> self.onfetch = function(event) {
>  event.respondWith(log(event.request).then(fetch));
> };

in

> .then(fetch)

What is this code doing ?
I assume this code is request to "api/quotations",right?

marco-c:
`log(event.request)` logs the request and returns a Promise that resolves to the request.
Once the promise is resolved, `fetch` is called with the request.

A more verbose equivalent would be:
```JavaScript
log(event.request)
.then(function(request) {
  return fetch(request);
})
```

PR #185, opened 2015-12-27
Clarify fetch handler in the api-analytics recipe

marco-c:
This PR helps to clarify the `fetch` handler of the api-analytics recipe (see #184).

### DIFF
--- a/api-analytics/service-worker.js
+++ b/api-analytics/service-worker.js
@@ -11,19 +11,31 @@ self.onactivate = function(event) {
   event.waitUntil(self.clients.claim());
 };
 
-// Fetch is as simply as log the request and passthrough.
-// Water clear thanks to the promise syntax!
 self.onfetch = function(event) {
-  event.respondWith(log(event.request).then(fetch));
+
+  event.respondWith(
+    // Log the request ...
+    log(event.request)
+    // .. and then actually perform it.
+    .then(fetch)
+  );
 };
 
 // Post basic information of the request to a backend for historical purposes.
 function log(request) {
-  var returnRequest = function() { return Promise.resolve(request); };
-  var data = { method: request.method, url: request.url };
+  var returnRequest = function() {
+    return request;
+  };
+
+  var data = {
+    method: request.method,
+    url: request.url
+  };
+
   return fetch(LOG_ENDPOINT, {
     method: 'POST',
     body: JSON.stringify(data),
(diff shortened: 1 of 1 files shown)


api-analytics/service-worker.js
```diff
@@ -11,19 +11,30 @@ self.onactivate = function(event) {
...
 };
 
-// Fetch is as simply as log the request and passthrough.
-// Water clear thanks to the promise syntax!
 self.onfetch = function(event) {
-  event.respondWith(log(event.request).then(fetch));
+  event.respondWith(
+    // Log the request ...
+    log(event.request)
+    // .. and then actually perform it.
```
delapuente: A dot is missing here.

api-analytics/service-worker.js
```diff
@@ -11,19 +11,30 @@ self.onactivate = function(event) {
...
+    .then(fetch)
+  );
 };
 
 // Post basic information of the request to a backend for historical purposes.
 function log(request) {
-  var returnRequest = function() { return Promise.resolve(request); };
-  var data = { method: request.method, url: request.url };
+  var returnRequest = function() {
+    return request;
```
delapuente: Despite I know here it's not necessary, I prefer the form returning the Promise because of two main reasons:
  1. It saves the normalization step.
  2. It's more versatile, allowing it to appear at the beginning (`returnRequest().then(...);`) or in the middle of the chain (`.then(returnRequest).then(...);`)

Not blocking on this.

marco-c: I prefer the form that doesn't return the Promise for synchronous functions.
It's true that it can't appear at the beginning, but since it's synchronous you don't need to do `returnRequest().then`, you can just call it and get the return value in a local variable.
It can appear in the middle of the chain, for example try running this script in the console:
```JavaScript
var str = 'marco';
function returnStr() {
  return str;
}
Promise.resolve().then(returnStr).then(console.log.bind(console));
```

delapuente: Of course, it can appear in the middle of the chain due to reason 1. The advantage of the other form is that it can appear in **both** places without altering. The good thing with promises, is that they are value abstractions. Wrapping operations in promise-returning functions allows you to hide synchronous / asynchronous details, working in an homogeneous way. The same can not be said for synchronous programming but... as said, not blocking on this.

delapuente:
Thanks for the enhancements.

PR #186, opened 2015-12-29
Add recipe that uses the service worker as a notification dispatcher to multiple tabs

marco-c:
I'd need help to choose the name :smile:


darkwing:
"Notification Dispatcher" ?

darkwing:
Wow, this is simple and wonderful! :)

I have two suggestions:

-  This currently doesn't work in Chrome, so we may want to add a note about that in the demo or in the README

-  It would be awesome to show "feedback" after they click the submit button, like "Now check your other tabs to see the message"

darkwing:
I might also suggest using a `<form>` tag so the user can hit the `ENTER` key to submit the message.

digitarald:
@marco-c I really like it, could be a good use case for apps (and +1 on David's feedback).

PR #191, opened 2016-01-08
Make design softer, less harsh

darkwing:
Last time I do this, I promise.  
screenshot at jan 08 09-51-41

### DIFF
--- a/src/css/style.css
+++ b/src/css/style.css
@@ -47,10 +47,6 @@ html {
   height: 100%;
 }
 
-a {
-  color: #a87aad;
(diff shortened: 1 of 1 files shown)


digitarald:
To keep it consistent we could we use the colors from Martell's new Mozilla ID palette: http://blog.seanmartell.com/2014/09/17/mozilla-id-project-palette-explorations/

Feel free to rip them from platatus' variables https://github.com/mozilla/platatus/blob/master/src/css/variables/mozid.css

PR #199, opened 2016-01-20
Add recipe that allows getting a payload without using PushMessageData

marco-c:


marco-c:
This would be useful when replying to some StackOverflow questions (since usually people want a browser-agnostic way to get the payload and Chrome doesn't support push payloads yet).

digitarald:
I like it, well documented!

PR #203, opened 2016-01-23
Update eslint-config-airbnb to version 4

marco-c:
I've updated eslint-config-airbnb, modifying some of the rules they added with what I prefer.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -7,6 +7,9 @@
     "vars-on-top": 0,
     "no-use-before-define": 0,
     "space-before-function-paren": 0,
-    "max-len": [1, 80, 4, {"ignoreComments": true, "ignoreUrls": true}]
+    "max-len": [1, 80, 4, {"ignoreComments": true, "ignoreUrls": true}],
+    "no-param-reassign": 0,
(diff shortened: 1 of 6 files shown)


.eslintrc
```diff
@@ -7,6 +7,9 @@
     "vars-on-top": 0,
     "no-use-before-define": 0,
     "space-before-function-paren": 0,
-    "max-len": [1, 80, 4, {"ignoreComments": true, "ignoreUrls": true}]
+    "max-len": [1, 80, 4, {"ignoreComments": true, "ignoreUrls": true}],
+    "no-param-reassign": 0,
```
delapuente: OMG! I was wondering why we have this option enabled at all!

.eslintrc
```diff
@@ -7,6 +7,9 @@
...
+    "quote-props": 0,
+    "wrap-iife": [2, "inside"]
```
delapuente: Just asking. Is there some reason for this or is it just a matter of preference?

marco-c: Just a matter of preference, but I don't really care, we can disable it if you prefer the other way.

PR #207, opened 2016-02-15
Run lint when building

marco-c:
I use `gulp start-server` while developing, which depends on `build`. `build` doesn't depend on `test` and on `lint` though, so I often forget to run `lint`.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -204,7 +204,7 @@ gulp.task('build:favicon', ['clean'], function() {
 });
 
 // Start express server after building site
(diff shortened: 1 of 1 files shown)


TimAbraldes:
Often when I build I've inserted test code that won't pass lint so I find it a little cumbersome for build to always call lint. Would it be an acceptable solution to have Travis run lint on every push?

marco-c:
@TimAbraldes I've updated the PR to run `lint` only when you run `start-server`.

marco-c:
We've discussed this on IRC, and @TimAbraldes was OK with it.

Issue #218, opened 2016-03-24
Add recipe with VAPID


marco-c:

digitarald:
@groovecoder is looking at that

aliams:
👍 for this

slokhorst:
Fixed in #294

PR #219, opened 2016-03-24
Update web-push dependency

marco-c:
This adds support for Chrome 50+ to the `push-payload` and the `push-quota` recipes. Also, it's needed for VAPID (#218).

### DIFF
--- a/package.json
+++ b/package.json
@@ -53,7 +53,7 @@
     "glob": "^7.0.0",
     "request": "^2.64.0",
     "swig": "^1.4.2",
-    "web-push": "^1.0.1"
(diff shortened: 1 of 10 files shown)


marco-c:
The update breaks compatibility with Firefox Nightly, we have to wait for https://bugzilla.mozilla.org/show_bug.cgi?id=1257821 to be fixed.

marco-c:
The needed patch in that bug has landed, so we can do this now.

marco-c:
I'm going to merge it, so we can support Chrome 50.
@zalun, can you review it? I'll create a new PR with other changes if necessary.

Issue #202, opened 2016-01-22
Generate landing page for categories


digitarald:
As a user I want to link to web push examples.

PR #221, opened 2016-03-29
Add category landing pages

darkwing:
Fixes https://github.com/mozilla/serviceworker-cookbook/issues/202

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -24,7 +24,7 @@ var recipeSlugs = glob.sync('./!(dist|node_modules|src|_recipe_template)/').map(
 var srcRecipes = recipeSlugs.map(function makePath(name) {
   return './' + name + '/**';
 });
-var recipes = parseRecipes(recipeSlugs);
+var recipes = parseRecipes.parse(recipeSlugs);
 
 var template = (function() {
   function renderContent(content, options) {
@@ -35,6 +35,7 @@ var template = (function() {
       recipes: recipes,
       ghBase: 'https://github.com/digitarald/serviceworker-cookbook/blob/master/',
       currentRecipe: newOptions.currentRecipe,
+      categories: parseRecipes.categories
     });
   }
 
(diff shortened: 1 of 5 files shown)


digitarald:
Maybe a follow up bug, but could we have a description per category page for better SEO and giving users a primer on the technology.

parseRecipes.js
```diff
@@ -9,9 +9,19 @@ var difficulties = {
   'Beginner': 1
 };
 
-var categoryOrder = ['General Usage', 'Offline', 'Beyond Offline', 'Performance', 'Web Push'];
+var categories = module.exports.categories = [
+  { title: 'General Usage', slug: 'general-usage' },
```
digitarald: Should we include `service-workers` and `offline` for some SEO love?

darkwing: I don't think we need it to be honest -- those keywords are everywhere in each demo.

parseRecipes.js
```diff
@@ -9,9 +9,19 @@ var difficulties = {
...
+  { title: 'Offline', slug: 'offline' },
+  { title: 'Beyond Offline', slug: 'beyond-offline' },
+  { title: 'Performance', slug: 'performance' },
+  { title: 'Web Push', slug: 'web-push' }
+];
 
-module.exports = function(recipeSlugs) { 
+var categoryOrder = categories.map(function(category) {
+  return category.title;
+});
```
marco-c: Could we order them as before? (simple to hard, then `performance` and then `web push`)

marco-c: Ah, forget it, it's the same order.

parseRecipes.js
digitarald: In the demo but rarely in titles and headers so far. What about adding it to at least one section to see if it helps ranking.

gulpfile.js
```diff
@@ -129,6 +130,27 @@ gulp.task('build:index', ['clean'], function buildIndex() {
...
+      return recipe.category === category.title;
+    }).map(function(recipe) {
+      return {
+        slug: recipe.slug,
+        title: recipe.name,
+        summary: marked(recipe.summary)
+      };
+    });
+
+    return renderFile('./src/tpl/category.html', {
```
digitarald: I am assuming that we are setting the page title? I would expect something along the lines of "Beyond Offline Recipes - ServiceWorker Cookbook" for categories and "API Analytics - Beyond Offline - Service Worker Cookbook" for recipes.

marco-c: We're not, I'm going to file an issue for that.

marco-c: Filed #226.

PR #241, opened 2016-07-29
Delete the Chrome Warnings

dalmaer:
I noticed that the example worked on Chrome so I took a peak and it appears that it is fixed as of Chrome M51:

https://bugs.chromium.org/p/chromium/issues/detail?id=543198


marco-c:
Thanks!

PR #244, opened 2016-09-17
Polish

delapuente:
Make the cookbook compatible with the latest version of Firefox **stable** and fix the wrong calculus of the height of the iframe for those demos dynamically changing their document's height.

### DIFF
--- a/fetching/service-worker.js
+++ b/fetching/service-worker.js
@@ -3,7 +3,7 @@
 // Create a proxy for all requests to the local urls containing a
 // `cookbook-proxy` string.
 self.onfetch = function(event) {
-  if (event.request.url.contains('cookbook-proxy')) {
+  if (event.request.url.match('cookbook-proxy')) {
     var init = { method: 'GET',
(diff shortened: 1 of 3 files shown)


fetching/service-worker.js
```diff
@@ -3,7 +3,7 @@
 // Create a proxy for all requests to the local urls containing a
 // `cookbook-proxy` string.
 self.onfetch = function(event) {
-  if (event.request.url.contains('cookbook-proxy')) {
+  if (event.request.url.match('cookbook-proxy')) {
```
marco-c: [contains][1] has been removed in Firefox 48 and replaced by [includes][2]. So you could use `includes` instead of `match`.

[1]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/includes#String.prototype.contains
[2]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/includes

PR #248, opened 2016-09-19
Adding common assets for offline strategies recipes

delapuente:
This is the root for all cache strategies recipes. The assets are needed to avoid delays introduced by proxing from other sites.

@marco-c or @potch do you mind to take a quick look?

PR #249, opened 2016-09-19
Offline fallback strategy

delapuente:


marco-c:
Can you update the PRs now that #248 has been merged?

strategy-offline-fallback/README.md
```diff
@@ -0,0 +1,17 @@
+# Offline fallback
+The recipe provides a service worker serving a default offline fallback in case
+of missing resource.
+
+## Difficulty
+Intermediate
+
+## Use Case
+You want to fail in a controlled way while serving resources providing content
```
sole: I think I'd say something like "you want to make sure the users always receive some content, even if the network is not available"

strategy-offline-fallback/controlled.html
```diff
@@ -0,0 +1,21 @@
...
+  <style>
+   body {
+     text-align: center;
+   }
+  </style>
+</head>
+<body>
+  <h1>Offline fallback</h1>
+  <img src="" alt="missing asset" />
+  <p>This image request points to an unavailable resource but since the page
```
sole: Add some commas so we can breathe while reading this!

[...] resource**,** [...] fallback**,**

strategy-offline-fallback/controlled.js
```diff
@@ -0,0 +1,18 @@
+// To be able to provide fallbacks since the beginning we can only load the
```
sole: from the beginning

strategy-offline-fallback/controlled.js
```diff
@@ -0,0 +1,18 @@
...
+// image once we know the SW is ready. In addition, the SW should start
+// intercepting ASAP.
```
sole: Please don't use abbreviations in the comments unless necessary. This particular block is littered with acronyms and makes things hard to read

strategy-offline-fallback/controlled.js
```diff
@@ -0,0 +1,18 @@
...
+if (navigator.serviceWorker.controller) {
```
sole: why are you checking for the `controller` property? how does it work? this is a bit confusing

strategy-offline-fallback/non-controlled.html
```diff
@@ -0,0 +1,19 @@
...
+  <title>Cache only: non controlled page - ServiceWorker Cookbook</title>
+  <meta name="viewport" content="width=device-width, initial-scale=1">
+  <style>
+   body {
+     text-align: center;
+   }
+  </style>
+</head>
+<body>
+  <h1>Non controlled, no fallback</h1>
```
sole: Non -> not

strategy-offline-fallback/service-worker.js
```diff
@@ -0,0 +1,71 @@
+var CACHE = 'offline-fallback';
+
+// On install, cache the non available resource.
+self.addEventListener('install', function(evt) {
+  console.log('The service worker is being installed.');
+  evt.waitUntil(precache().then(function () {
+    // Skip waiting is necessary since we want the SW to be active ASAP
```
sole: since -> because

stop using ASAP

add a comma before instead

strategy-offline-fallback/service-worker.js
```diff
@@ -0,0 +1,71 @@
...
+    // instead of waiting for non controlled clients to be closed first.
+    return self.skipWaiting();
+  }));
+});
+
+self.addEventListener('activate', function (evt) {
+  // `self.clients.claim()` allows the service worker to start intercepting
+  // requests immediately. In addition to `self.skipWaiting()` is needed to
```
sole: I think you're missing an 'it' before 'is needed' ?

strategy-offline-fallback/service-worker.js
```diff
@@ -0,0 +1,71 @@
...
+  // allow serving fallbacks since the beginning.
```
sole: since -> from

strategy-offline-fallback/service-worker.js
```diff
@@ -0,0 +1,71 @@
...
+  evt.waitUntil(self.clients.claim());
+});
+
+// Use any strategy you want but check for errors to finally use the fallback.
```
sole: comma after `want`

I don't understand the rest of the comment. Do you mean the developer must check for errors before using the fallback? Or that they must check for errors and use the fallback in that case?

strategy-offline-fallback/service-worker.js
```diff
@@ -0,0 +1,71 @@
...
+  return caches.open(CACHE).then(function (cache) {
+    return cache.addAll([
+      './controlled.html'
+    ]);
+  });
+}
+
+// This is a simplified version of network then cache without timeout to
+// illustrate error control.
+function whateverStrategy(request) {
```
sole: this is probably not a very fortunate name, as it doesn't describe what is going on and it's also distracting me from focusing on the code to understand what it does and come up with a suggestion for a better name :-/

strategy-offline-fallback/service-worker.js
```diff
@@ -0,0 +1,71 @@
...
+function useFallback() {
+  return Promise.resolve(new Response(FALLBACK, { headers: {
+    'Content-Type': 'image/svg+xml'
+  }}));
+}
+
+function fromCache(request) {
+  return caches.open(CACHE).then(function (cache) {
+    return cache.match(request).then(function (matching) {
+      return matching || Promise.reject('no-match');
```
sole: what's this mysterious `no-match`? where does it come from? what does it do?

PR #251, opened 2016-09-19
Cache only strategy

delapuente:


strategy-cache-only/README.md
```diff
@@ -0,0 +1,16 @@
...
+## Difficulty
+Beginner
+
+## Use Case
+For a given version of your site, you have static content that never changes
+such as the shell around the content.
+
+## Solution
+Add static content during the installation of the service worker and use the
+cache to retrieve it whether there were network or not.
```
sole: I think "whether the network is available or not" would sound a bit more idiomatic

strategy-cache-only/controlled.html
```diff
@@ -0,0 +1,20 @@
...
+  <style>
+   body {
+     text-align: center;
+   }
+  </style>
+</head>
+<body>
+  <h1>Cache only</h1>
+  <img src="./asset" alt="sample asset" />
+  <p>This image request originates from a controlled page so the image will
```
sole: What does a "controlled page" mean? Is it a page under the service worker scope? (maybe it'd be better to describe it like that?)

strategy-cache-only/index.js
```diff
@@ -0,0 +1,45 @@
+// Register the ServiceWorker limiting its action to those URL starting
```
sole: what about "limiting its scope of action", or "limiting its scope"?

strategy-cache-only/non-controlled.html
```diff
@@ -0,0 +1,18 @@
...
+  <title>Cache only: non controlled page - ServiceWorker Cookbook</title>
+  <meta name="viewport" content="width=device-width, initial-scale=1">
+  <style>
+   body {
+     text-align: center;
+   }
+  </style>
+</head>
+<body>
+  <h1>Always synchronized</h1>
```
sole: I'm not sure 'synchronized' works here, perhaps 'always fresh from the server' would be more clear. Synchronized makes me think of synchronising content both ways and that's not what's happening...

strategy-cache-only/service-worker.js
```diff
@@ -0,0 +1,34 @@
+var CACHE = 'cache-only';
+
+// On install, cache some resource.
```
sole: some resource**s**

strategy-cache-only/service-worker.js
```diff
@@ -0,0 +1,34 @@
...
+});
+
+// On fetch, use cache only strategy.
+self.addEventListener('fetch', function(evt) {
+  console.log('The service worker is serving the asset.');
+  evt.respondWith(fromCache(evt.request));
+});
+
+// Open a cache and use `addAll()` with an array of assets to add all of them
+// to the cache. Extend the installation until all the assets are saved.
```
sole: what does 'extend the installation' mean? do you mean the installation isn't finished until the assets are saved to the cache?

strategy-cache-only/service-worker.js
```diff
@@ -0,0 +1,34 @@
...
+function precache() {
+  return caches.open(CACHE).then(function (cache) {
+    return cache.addAll([
+      './controlled.html',
+      './asset'
+    ]);
+  });
+}
+
+// As simple as opening the proper cache and search for the requested resource.
```
sole: "simple" is very relative. I'd suggest rewriting this comment as something like 'We will open the cache where we stored the data before, and search for the requested resource'

strategy-cache-only/service-worker.js
```diff
@@ -0,0 +1,34 @@
...
+// Notice in case of no matching, undefined is passed.
```
sole: 'returned', rather than 'passed'

delapuente:
@sole done. I though your comments will dissapear or shown as obsolete but they did not. Anyway, the las commit made the changes.

strategy-cache-only/index.html
```diff
@@ -0,0 +1,42 @@
...
+   }
+  </style>
+</head>
+<body>
+  <h1>Cache only</h1>
+  <div id="comparison">
+    <iframe src="./non-controlled.html" id="reference"></iframe>
+    <iframe src="./controlled.html" id="sample"></iframe>
+  </div>
+  <p>The images in these iframes points to the same asset in the server. But the first is controlled by the service worker and the second is not.</p>
```
sole: *point

strategy-cache-only/index.js
```diff
@@ -0,0 +1,45 @@
...
+  scope: './controlled'
+});
+
+// Load controlled and uncontrolled pages once the worker is active.
+navigator.serviceWorker.ready.then(reload);
+
+var referenceIframe = document.getElementById('reference');
+var sampleIframe = document.getElementById('sample');
+
+// Fix heights every time the iframe reload.
```
sole: *"is reloaded" or "reloads" ?

PR #278, opened 2017-02-27
Removed warning about Chrome is message relay

### LINKED ISSUE #141: Remove warning from message-relay recipe once it works in Chrome
We should remove the warning added in #140 once the recipe works in Chrome.

delapuente:
Fix #141

### DIFF
--- a/message-relay/service-worker.js
+++ b/message-relay/service-worker.js
@@ -24,8 +24,8 @@ self.addEventListener('message', function(event) {
     });
   });
 
-  // If event.waitUntil is defined (not yet in Chrome because of the same issue detailed before),
-  // use it to extend the lifetime of the Service Worker.
+  // If event.waitUntil is defined, use it to extend the
+  // lifetime of the Service Worker.
   if (event.waitUntil) {
     event.waitUntil(promise);
   }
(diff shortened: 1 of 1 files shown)


message-relay/service-worker.js
```diff
@@ -4,9 +4,6 @@ self.addEventListener('message', function(event) {
   var promise = self.clients.matchAll()
   .then(function(clientList) {
     // event.source.id contains the ID of the sender of the message.
-    // `event` in Chrome isn't an ExtendableMessageEvent yet (https://slightlyoff.github.io/ServiceWorker/spec/service_worker/#extendablemessage-event-interface),
-    // so it doesn't have the `source` property.
-    // https://code.google.com/p/chromium/issues/detail?id=543198
     var senderID = event.source ? event.source.id : 'unknown';
```
marco-c: We should remove the if as well, since now `event.source` is always available.

message-relay/service-worker.js
```diff
@@ -4,9 +4,6 @@ self.addEventListener('message', function(event) {
...
 
     // We'll also print a warning, so users playing with the demo aren't confused.
```
marco-c: And we should also remove this warning.

PR #292, opened 2018-01-22
Allow startup without setting GCM API key

dmonego:
Noticed that this was blocking startup, so made a fix that seemed to be in line with recent progress on this issue.


marco-c:
Thanks!

PR #293, opened 2018-01-22
Use .webmanifest as extension for Web App Manifests

slokhorst:
As noted in specification: https://www.w3.org/TR/appmanifest/


marco-c:
It looks good to me.

Issue #290, opened 2018-01-16
Move push demos to VAPID


delapuente:

slokhorst:
I'm trying to do this, but it requires storing the server's VAPID keys somewhere. (like how the GCM key is stored in `manifest.json` right now)

What is the best way to do this? Do we already depend on some kind of node.js storage library?

The alternative would be to invalidate the subscription (client-side) on every page load and create a new one. But that's not good practice to teach people and wasteful because we are creating and destroying subscriptions all the time.

slokhorst:
There's a caveat here: If we switch to VAPID (remove GCM key), Chrome doesn't accept unencrypted push messages. Even when they have no data. This means that all push recipes will need the key handling/encryption code, even push-simple and push-get-payload. This makes them pretty much identical to push-payload, and they should probably be removed.

slokhorst:
Fixed by #294

PR #294, opened 2018-01-28
Convert all push recipes to use VAPID

slokhorst:
This is a start for implementing #290

Mostly adapted the example from https://github.com/web-push-libs/web-push

Some questions/notes before I do the other push demos:
* `urlBase64ToUint8Array` is a big and ugly, and will be needed in every push example. Should this be moved to a separate library file?
* Is there a way to compare `Uint8Array`s without using `toString()`?
* I've removed the GCM functionality as it's non-standard and should not be encouraged. This means we drop support for Chrome 51 and earlier. (almost 2 years old so should be ok)


push-payload/index.js
```diff
@@ -2,20 +2,51 @@ var endpoint;
...
   // Use the PushManager to get the user's subscription to the push service.
   return registration.pushManager.getSubscription()
-  .then(function(subscription) {
-    // If a subscription was found, return it.
+  .then(async function(subscription) {
+    // Get the server's public key
+    var response = await fetch('./vapidPublicKey')
+    var vapidPublicKey = await response.text();
+    // Convert the URL safe base64 string to a Uint8Array
+    var convertedVapidKey = urlBase64ToUint8Array(vapidPublicKey);
```
marco-c: You can probably use the base64 string directly, see https://developer.mozilla.org/en-US/docs/Web/API/PushManager/subscribe

push-payload/index.js
```diff
@@ -2,20 +2,51 @@ var endpoint;
...
+
+    // See if we already have a subscription with the server
     if (subscription) {
-      return subscription;
+      var currentKey = new Uint8Array(subscription.options.applicationServerKey);
```
marco-c: I'm not sure that we need to take into account the possibility of the vapid key changing in these examples.

marco-c:
> I've removed the GCM functionality as it's non-standard and should not be encouraged. This means we drop support for Chrome 51 and earlier. (almost 2 years old so should be ok)

This is fine, we are going to remove it from the `web-push` library too soon.

push-payload/index.js
slokhorst: If we don't store the key somewhere (as the example is right now), it will change every time the server process restarts. When the key has changed the example will no longer work: `server.js` will attempt to send the push message and the endpoint will return an 400 UnauthorizedRegistration response (Google) or 401 Unauthorized response with "Request did not validate missing authorization header" (Mozilla). 

This obviously happens a lot when developing locally. I'm not sure what kind of infrastructure serviceworke.rs is running on, but say you try the example one day, the server restarts overnight, and you try it again the following day, it will no longer work.

push-payload/index.js
slokhorst: Doesn't seem to work on Chrome:
`TypeError: Failed to execute 'subscribe' on 'PushManager': The provided value is not of type '(ArrayBuffer or ArrayBufferView)'`

marco-c: OK, so I guess it's not implemented yet in Chrome. Unfortunately I don't think there's a way to share the code between the recipes, so we will have to copy it everywhere.
Please also add a comment explaining why it is needed and that it can be removed once Chrome supports passing a base64 string.

push-payload/index.js
marco-c: We are using Heroku. I think we can store the VAPID key in a file so it doesn't get recreated every time.
It will still be recreated from time to time, but not every time you reload.

marco-c: It is still going to be a pain though... Hopefully we can figure out a better solution.

push-payload/index.js
slokhorst: Related [Chrome bug](https://bugs.chromium.org/p/chromium/issues/detail?id=802280)

push-payload/index.js
slokhorst: I still think the current check is a pretty good solution. Even in production scenarios you might need to change the keys sometimes (when the key is compromised or lost). This seems a sensible place to check if the subscription we have is still usable. And as a bonus we don't have to extend the demo with persistent storage (or something behind-the-scenes specifically for serviceworke.rs).

slokhorst:
By the way, if we no longer use GCM, the `manifest.webmanifest` is no longer required. Should I remove it?

marco-c:
> By the way, if we no longer use GCM, the manifest.webmanifest is no longer required. Should I remove it?

Yes, we can remove it as part of this PR.

slokhorst:
5f78284 makes things a lot more readable (from [web-push-libs](https://github.com/web-push-libs/web-push): "The expected format is the same output as JSON.stringify'ing a PushSubscription in the browser.")

If there aren't any more concerns (except the handling of changing server keys), I'll continue with the other recipes.

push-payload/index.js
```diff
@@ -1,34 +1,57 @@
...
-        btoa(String.fromCharCode.apply(null, new Uint8Array(rawKey))) :
-        '';
-  var rawAuthSecret = subscription.getKey ? subscription.getKey('auth') : '';
-  authSecret = rawAuthSecret ?
-               btoa(String.fromCharCode.apply(null, new Uint8Array(rawAuthSecret))) :
-               '';
-
-  endpoint = subscription.endpoint;
+  // Keep the subscription for easy access
+  mySubscription = subscription;
```
marco-c: We shouldn't need this, we can use `subscription` directly.

slokhorst: It's reused at line 63 in the doIt.onclick function. If we don't serve it here we need to add the following there (fine with me either way, but this adds some more code):

```javascript
navigator.serviceWorker.ready.then(function(registration) {
    registration.pushManager.getSubscription().then(async function(subscription) {
        ...
    }
}
````

File: push-payload/index.js (hunk 2 above)
Comment (marco-c) [reply push-payload/index.js]: I have another idea. We can put the VAPID key in a Heroku environment variable, so it will never change.

File: push-payload/index.js (hunk 3 above)
Comment (marco-c) [reply push-payload/index.js]: We can bring doIt.onclick in the scope here, it needs to wait for the subscription anyway.

File: push-payload/index.js [hunk 4]
```diff
@@ -1,69 +1,84 @@
-var endpoint;
-var key;
-var authSecret;
+// This function is needed because Chrome doesn't accept a base64 encoded string
+// as value for applicationServerKey in pushManager.subscribe yet
+// https://bugs.chromium.org/p/chromium/issues/detail?id=802280
+function urlBase64ToUint8Array(base64String) {
```
marco-c: Maybe we can put this function in a utils.js script in the top-directory, which we include in each recipe before index.js. What do you say?

slokhorst: That seems like the cleanest solution.. Unfortunately I'm not sure how to do this. If we put the file in the top-directory, it isn't served by the server process. I think it has to be specified in `gulpfile.js` to be served? Also not ideal.. Is there another way?

marco-c: Yes, probably the `gulpfile.js` has to copy the file to `dist/`.
@delapuente any other idea how to do this?

slokhorst: 1620dfc works, but might not be the best solution.

marco-c: That looks sane enough, maybe move it to the top repo directory though instead of src/js/.
I'm still undecided whether the best solution is to have the function in a separate file or not. One one hand, it's a pain to copy and paste it in every recipe, on the other hand, the recipes are supposed to be self-contained.

marco-c: (Sorry if the discussion is taking long, but once we have decided here we will apply the same rules to all recipes, so it will be way faster then).

slokhorst: Yep I agree that it's both not optimal.. Its a shame that Chrome hasn't even started their native implementation yet.. otherwise we could maybe leave the function out completely. But since it's not even on the horizon that's not really a possibility either.

(I've moved tools.js to the top directory in 1bb1db5)

slokhorst:
So what's the next step here? I'm not sure how to handle the key with a Heroku variable (will that even work locally on my machine?). Maybe it would be easier if you do that.

marco-c:
> So what's the next step here? I'm not sure how to handle the key with a Heroku variable (will that even work locally on my machine?). Maybe it would be easier if you do that.

As far as the application is concerned, a Heroku variable is just an environment variable. So you can emulate everything locally by defining an env variable.

slokhorst:
Ok, done in 68aebbc.

I'm still not convinced that removing the public key check in the browser is a good idea. If the server's key is (accidentally) changed it is not clear what's the problem (the `webPush.sendNotification()` call will get a 400 response an that's it). And then the only way to fix it is to go into the browser and manually unsubscribe:
```javascript
navigator.serviceWorker.getRegistration()
.then(registration => registration.pushManager.getSubscription())
.then(subscription => subscription.unsubscribe());
```

push-payload/index.js
```diff
@@ -1,69 +1,61 @@
...
+    // urlBase64ToUint8Array() is defined in /tools.js
+    var convertedVapidKey = urlBase64ToUint8Array(vapidPublicKey);
+
+    // See if we already have a subscription with the server
     if (subscription) {
       return subscription;
     }
 
-    // Otherwise, subscribe the user (userVisibleOnly allows to specify that we don't plan to
+    // Subscribe the user (userVisibleOnly allows to specify that we don't plan to
```
marco-c: Nit: don't change this comment.

push-payload/index.js
```diff
@@ -1,69 +1,61 @@
...
+    // Get the server's public key
+    var response = await fetch('./vapidPublicKey')
+    var vapidPublicKey = await response.text();
+    
+    // Chrome doesn't accept the base64-encoded (string) vapidPublicKey yet
+    // urlBase64ToUint8Array() is defined in /tools.js
+    var convertedVapidKey = urlBase64ToUint8Array(vapidPublicKey);
+
+    // See if we already have a subscription with the server
     if (subscription) {
```
marco-c: Move this at the beginning of the function.

push-payload/index.js
```diff
@@ -1,69 +1,61 @@
...
 // Register a Service Worker.
 navigator.serviceWorker.register('service-worker.js')
 .then(function(registration) {
   // Use the PushManager to get the user's subscription to the push service.
   return registration.pushManager.getSubscription()
-  .then(function(subscription) {
-    // If a subscription was found, return it.
+  .then(async function(subscription) {
+    // Get the server's public key
+    var response = await fetch('./vapidPublicKey')
```
marco-c: Nit: use `let` and `const` everywhere instead of `var` when possible.

marco-c:
LGTM overall, I think we can convert the other recipes too now.

slokhorst:
Converted all recipes and rebased.

I think we should remove push-simple, as it's pretty much identical to push-payload now, see https://github.com/mozilla/serviceworker-cookbook/issues/290#issuecomment-369330773.

marco-c:
So many changes!
Overall it looks good, I need to set the Heroku key before merging.

marco-c:
@delapuente do you want to take a look too?

delapuente:
I'll do. Moved to Windows environment and I also need to set everything up again.

push-get-payload/server.js
```diff
@@ -2,25 +2,43 @@
...
+      TTL: req.body.ttl
+    };
+
     setTimeout(function() {
-      payloads[req.body.endpoint] = req.body.payload;
-      webPush.sendNotification({
-        endpoint: req.body.endpoint,
-        TTL: req.body.ttl,
-      })
+      payloads[req.body.endpoint] = payload;
```
delapuente: You need to index the `payloads` map by the endpoint which is now [*inside* the `subscription` object](https://github.com/mozilla/serviceworker-cookbook/pull/294/files#diff-12d95901e9b60d3002f4388bbc702b06R52): 
```js
payloads[req.body.subscription.endpoint] = payload;
```

slokhorst: Oops! Strange I didn't notice this. Fixed.

marco-c:
I've generated VAPID keys and set the two env variables on Heroku.

push-clients/server.js
```diff
@@ -2,22 +2,40 @@
...
+  console.log("You must set the VAPID_PUBLIC_KEY and VAPID_PRIVATE_KEY "+
+    "environment variables. You can use the following ones:");
+  console.log(webPush.generateVAPIDKeys());
+  return;
+}
+// Set the keys used for encrypting the push messages.
+webPush.setVapidDetails(
+  'https://serviceworke.rs/',
+  process.env.VAPID_PUBLIC_KEY,
+  process.env.VAPID_PRIVATE_KEY,
```
marco-c: Heroku is failing because of the comma here.

aliams:
Looking over the changes, I see that subscribe() is called when the service worker registration promise resolves. Considering that the service worker may not actually be ready at that point, should we change it so that the subscribe() is called when navigator.serviceWorker.ready resolves?

aliams:
Thanks for the extra context. Because the ready promise is meant to be resolved when the registration has an active service worker and subscribe() requires an active service worker, I believe it would make sense to call subscribe when ready resolves. Currently, as it is, it will not work in insider builds of Edge, but if it's changed to use ready, it will work.

aliams:
Thank you @marco-c. I've just opened #302.

PR #303, opened 2018-04-08
Wait for serviceWorker.ready in push recipes

### LINKED ISSUE #302: Change samples to subscribe when serviceWorker.ready resolves
The code samples should be changed to only call the subscribe() function when the serviceWorker.ready promise resolves.

#170, though similar, discussed how to ensure the page was controlled by the service worker which required the use of the controllerchange event.

For push, we do not need to wait for the page to be controlled, but rather we just need the service worker registration to have an active service worker (i.e. when the serviceWorker.ready promises resolves).

slokhorst:
Fix #302

### DIFF
--- a/push-clients/server.js
+++ b/push-clients/server.js
@@ -28,14 +28,8 @@ module.exports = function(app, route) {
   });
 
   app.post(route + 'sendNotification', function(req, res) {
-    const subscription = req.body.subscription;
-    const payload = null;
-    const options = {
-      TTL: req.body.ttl
-    };
-
     setTimeout(function() {
-      webPush.sendNotification(subscription, payload, options)
+      webPush.sendNotification(req.body.subscription)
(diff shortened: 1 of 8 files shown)


push-subscription-management/index.js
```diff
@@ -1,34 +1,23 @@
-// [Working example](/push-subscription-management_demo.html).
```
marco-c: The changes in this recipe are not needed as part of this PR, right?
I'm not against them, but they should be made in a separate PR.

push-clients/server.js
```diff
@@ -31,7 +31,7 @@ module.exports = function(app, route) {
     const subscription = req.body.subscription;
     const payload = null;
     const options = {
```
marco-c: Let's remove these options completely and use the web-push library's default.

PR #304, opened 2018-04-12
Cleanup push-subscription-management

slokhorst:
Mostly cosmetic to get things in the same style as the other recipes.
It seems like the first few markdown lines are no longer used?


push-subscription-management/index.js
```diff
@@ -1,34 +1,23 @@
...
-        console.log('Already subscribed', subscription.endpoint);
-        setUnsubscribeButton();
-      } else {
-        setSubscribeButton();
-      }
-    });
-}
+// When the Service Worker is ready, enable the UI (button),
+// and see if we already have a subscription set up.
+navigator.serviceWorker.ready.then(function(registration) {
```
marco-c: Nit: put `.then` in a new line like the other recipes

marco-c: (here and in the following cases)

PR #334, opened 2021-12-23
Update server.js : Disable redirect to serviceworke.rs

mimsadAlef:
Due to the deactivation of the domain serviceworke.rs, I suggest disabling the redirect to open the site through serviceworker-cookbook.herokuapp.com.

### DIFF
--- a/server.js
+++ b/server.js
@@ -29,9 +29,9 @@ app.use(function forceLiveDomain(req, res, next) {
   // Don't allow user to hit Heroku now that we have a domain
   var host = req.get('Host');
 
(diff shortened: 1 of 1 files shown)


marco-c:
Unfortunately the build is failing and there is no one maintaining this project anymore. I'd be happy to accept this, but we'd need to fix the build otherwise the fix will never get deployed.

marco-c:
There are a few other instances of `serviceworke.rs` in the code to change, e.g https://github.com/mozilla/serviceworker-cookbook/blob/b3d7788d44c5c1f071b5321b5ee1c5d7dbcdfec5/push-rich/server.js#L15.

